Issue #4, opened 2015-01-06
Allow use of a custom class for user on audits


pjb3:
In our app, the User model isn't named User, so you get an error when trying to create a query.

ankane:
Fixed in #6.

PR #5, opened 2015-01-06
Allow custom user class name

pjb3:
Resolves issue #4

### DIFF
--- a/app/models/blazer/audit.rb
+++ b/app/models/blazer/audit.rb
@@ -1,6 +1,6 @@
 module Blazer
   class Audit < ActiveRecord::Base
-    belongs_to :user
+    belongs_to :user, class_name: ENV['BLAZER_USER_CLASS_NAME'] || DEFAULT_USER_CLASS_NAME
     belongs_to :query
   end
 end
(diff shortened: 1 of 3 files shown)


ankane:
Hey Paul, thanks for the pull request?  How about:

```ruby
Blazer.user_class_name = 'Admin'
```

to keep it consistent with the other customizations?

pjb3:
@ankane I wanted to do that too initially but I don't think that works because you have to reference the value when you define the associations and you won't have a chance to set that value in an initializer before the models in blazer are loaded.

paultyng:
Spree does it via an initializer and referencing the variable in its associations, so I think that works.

`Spree::Order` model: https://github.com/spree/spree/blob/master/core/app/models/spree/order.rb#L38

ankane:
@pjb3 does #6 work for you?

pjb3:
@ankane Yes, @paultyng's solution works fine, I didn't realize the Rails models get loaded after the initializers run

ankane:
Thanks @pjb3, just merged #6.

PR #6, opened 2015-01-14
Use initializer variable for user class name

paultyng:
This should address #4 and #5 

The initializer seems to work for me, and as far as I know I haven't customized my loading unless a gem in my app did it unknowingly.


ac21:
+1;  this would be really helpful

ankane:
Awesome, thanks Paul!

PR #7, opened 2015-01-24
provide extension capability to blazer controllers

ac21:
Adds the ability to extend blazer controllers.  Can be useful in adding callbacks and actions ("current_user")

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,6 +101,31 @@ authenticate :user, lambda{|user| user.admin? } do
 end
 ```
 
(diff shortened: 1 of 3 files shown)


ankane:
Hey @ac21, you can do:

```ruby
Blazer::QueriesController.send(:include, Blazer::Controller)
```

in an initializer with the current code.

ac21:
@ankane - thanks!   I'll do that instead.

PR #8, opened 2015-01-24
addresses mysql2 issue with empty column_types on select statements

ac21:
the mysql2 adapter currently can't be used with Blazer.  Seems to be an issue that mysql2 directly run sql statements `Blazer::Connection.select(statement)` returns empty column_types.   A simple patch to not type_cast the data when column_types are empty allows basic functionality

### DIFF
--- a/app/controllers/blazer/queries_controller.rb
+++ b/app/controllers/blazer/queries_controller.rb
@@ -160,7 +160,7 @@ module Blazer
           result.each do |untyped_row|
             row = {}
             untyped_row.each do |k, v|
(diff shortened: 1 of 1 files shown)


ankane:
Hey @ac21, what version of Rails are you on?

ac21:
Rails 4.1.6 using mysql2 0.3.17.   works fine using mysql adapter but mysql2 doesn't seem to return column_types

ankane:
Looks like the mysql2 adapter does the typecasting already.  Thanks for debugging and fixing this!  New release going out in a few.

PR #11, opened 2015-03-07
Dynamic columns

backus:
This pull request would add support for `dynamic_columns` which are similar to `smart_variables`. I'll demonstrate why `dynamic_columns` are helpful by sharing my use case:

### Use case

I have a `Name` model with six different text fields I care about:

```ruby
>> Name.inspect
=> "Name(id: integer, first: string, second: string, third: string, fourth: string, fifth: string, full: text)"
```

and I want to create a blazer page that counts the number of rows where a name column has a certain string length.  I could make a blazer page like this for each of the six fields:

image

But that is hardly ideal. Instead, I define a dynamic column in my `blazer.yml`:

```yml
dynamic_columns:
  name_attribute: SELECT column_name FROM information_schema.columns WHERE table_name='names' AND (data_type='character varying' OR data_type='text');
```

and this lets me boil those six blazer pages down to one:

image

### DIFF
--- a/app/controllers/blazer/queries_controller.rb
+++ b/app/controllers/blazer/queries_controller.rb
@@ -41,7 +41,7 @@ module Blazer
       @smart_vars = {}
       @sql_errors = []
       @bind_vars.each do |var|
-        query = smart_variables[var]
+        query = smart_variables[var] || dynamic_columns[var]
         if query
           rows, error = run_statement(query)
           @smart_vars[var] = rows.map{|v| v.values.reverse }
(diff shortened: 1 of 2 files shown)


backus:
@ankane Would you be open to a separate pull request that cleans up `Blazer::QueriesController` ([code climate evaluation](https://codeclimate.com/github/ankane/blazer/Blazer::QueriesController))? I would be happy to pull many of the methods into a few modules.

alainmeier:
This is pretty cool - any chance this will be merged into master, @ankane?

ankane:
Thanks for the pull request John! Really sorry for the delay. I'm a bit hesitant to introduce a new concept if we can avoid it. What do you think of using smart variables for this, but prevent escaping with `{{name_attribute}}`?

backus:
That sounds better. It has been a while since I wrote this, but I believe we could just configure it so `{attribute}` uses [`quote`](http://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/Quoting.html#method-i-quote) and `{{attribute}}` uses [`quote_column_name`](http://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/Quoting.html#method-i-quote_column_name) instead of just not escaping at all. Happy to modify the PR to do that.

ankane:
If we don't escape at all, this could also be used for:

```ruby
SELECT * FROM users WHERE name LIKE '%{{name}}%'
```

(one of the common problems I run into)

backus:
Sorry I forgot to followup. Wouldn't not escaping input be a bit problematic? I can imagine this unexpectedly breaking queries...

PR #15, opened 2015-05-15
Use secrets.yml instead of ENV directly

ameech:
Found it easier to use secrets.yml instead of trying to set ENV variables for all our settings.


ameech:
Wrong branch :(

PR #16, opened 2015-05-15
Use secrets.yml instead of ENV directly

ameech:
Found it easier to use secrets.yml instead of trying to set ENV variables for all our settings.


ankane:
Thanks for the pull request Adam!  However, I'd prefer to stick with `ENV` vars.  Check out [dotenv-rails](https://github.com/bkeepers/dotenv) for an easy way to manage them in development.

PR #17, opened 2015-05-26
Made the editor bigger.

far33d:
The editor is so small. 
Let's make it bigger.


starchow:
Voted :+1:

ankane:
Thanks for the pull request @far33d! The editor now expands as you type, so it's no longer needed.

PR #22, opened 2015-09-23
fix devise model name (example - administarator)

ermolaev:
for model `Administarator` devise generate method `current_administrator`, it not worked in blazer


ankane:
Great, thanks @ermolaev!  I changed the option name to `user_method`.

PR #27, opened 2015-10-12
fix for json_escape prior to versions 4.1 of rails

djburdick:
I put in the update from rails 4.1+ for json_escape that fixes the double quotes issue. In case you want to support versions < 4.1.

Should be benign since it just calls "super" for newer Rails versions.
Support for "Rails::VERSION::STRING" goes back to Rails 2.

### DIFF
--- a/app/helpers/blazer/base_helper.rb
+++ b/app/helpers/blazer/base_helper.rb
@@ -1,6 +1,6 @@
 module Blazer
   module BaseHelper
-    include JsonEscape
+    include BlazerJsonEscape
 
     def title(title = nil)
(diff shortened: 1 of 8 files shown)


ankane:
Thanks @djburdick!  Unfortunately, this will change all instances of `json_escape` in the app since all helpers are included by default.  A safer approach would be to replace all instances of `json_escape` with `blazer_json_escape` and define `blazer_json_escape` conditionally (alias `json_escape` if 4.1 or above, and define it otherwise).

djburdick:
Good point. will update

PR #28, opened 2015-10-13
Change patched method name to blazer_json_escape

djburdick:


djburdick:
going to rebase these. will add a new request

PR #29, opened 2015-10-13
fix for json_escape prior to versions 4.1 of rails

djburdick:
change name of json_escape patch method to blazer_json_escape
call json_escape not super


ankane:
Thanks @djburdick!  Planning to release a new version today so others can use this.

PR #30, opened 2015-10-24
Add invert option to fail if no results are returned

ryninho:


ankane:
:+1: Thanks Eric, I'll take a look when I have some time.

ankane:
:+1: Thanks again!

PR #33, opened 2015-11-09
Fixes link to Smart Columns example

crystalschang:


ankane:
Thanks @crystalschang! The users-by-occupation query is actually an example of smart variables (confusing I know - need better names), so I updated the link to point to the smart columns example.

crystalschang:
@ankane Ah okay, yeah I got them mixed up... I was intending to update the Smart Variables example, not the Smart Columns example. :)

ankane:
Yeah, looks like that's wrong too :-/ Feel free to submit another PR, or I'll fix when I have a chance.

Issue #35, opened 2015-11-24
ActionController::InvalidAuthenticityToken (ActionController::InvalidAuthenticityToken)


duboff:
For API only app, I get the error above. The culprit is this:

``` rb
protect_from_forgery with: :exception
```

In https://github.com/ankane/blazer/blob/master/app/controllers/blazer/base_controller.rb#L6

Why is blazer overwriting the default protect_from_forgery  from the ApplicationController. Doesn't seem like it should be the case. I reopened and removed the line and everything is working. Happy to submit a PR if this makes sense.

PS absolutely love Blazer, saves us a lot of headache.

ankane:
It probably makes sense to keep the application default, in which case you'd want to exclude `verify_authenticity_token` from the `skip_filter` list.

gen0cide-:
@ankane - Any hesitation accepting this PR? Would rather not be using a fork :)

ankane:
Yeah, my previous comment was never addressed.  Thinking about it again, I'm not sure it makes sense.  What CSRF protection method would you use instead?

duboff:
@ankane I am sorry for not following up on this but I did not quite understand what you meant in the previous comment. Would be very keen to reopen discussions. I would think Blazer should just respect the application default.

@gen0cide you don't necessarily need a fork. We just have a Blazer::BaseController with:
```
protect_from_forgery with: :null_session
```

ankane:
Hey @duboff, no worries. How are you doing authentication?

duboff:
We use Clearance gem for the admin pages which Blazer is a part of.

ankane:
I'll follow up over email since this involves security.

richardvenneman:
@duboff did you manage to get this to work? I tried everything from monkey patching to forking and removing the `protect_from_forgery` directive as well as inheriting from `ActionController::Base`. I always ended up with the following error:

> {"status":500,"error":"Internal Server Error","exception":"#\u003cActionView::Template::Error: undefined local variable or method `new_query_path' for #\u003c#\u003cClass:0x007fb3f86eb2d0\u003e:0x007fb3f6e95a00\u003e\u003e","traces":{"Application Trace":[{"id":88,"trace":"appsignal (2.1.0) lib/appsignal/rack/rails_instrumentation.rb:16:in `call'"}]

Really hope someone has gotten Blazer to work with rails-api!

duboff:
[...]

PR #36, opened 2015-11-25
Avoid overwriting default protect_from_forgery behaviour

duboff:
It makes sense to respect the default behaviour typically defined in the ApplicationController rather than overwrite it. In particular, in API-only apps this leads to https://github.com/ankane/blazer/issues/35

### DIFF
--- a/app/controllers/blazer/base_controller.rb
+++ b/app/controllers/blazer/base_controller.rb
@@ -3,8 +3,6 @@ module Blazer
     # skip all filters
     skip_filter *_process_action_callbacks.map(&:filter)
 
-    protect_from_forgery with: :exception
-
     if ENV["BLAZER_PASSWORD"]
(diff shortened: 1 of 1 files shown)


duboff:
@ankane any feedback on this?

buren:
I also have an "API-only" app and ran into this problem since I was subclassing `ActionController::API` in `ApplicationController`. What I ended up doing and actually I really liked the approach in the end was creating a new controller `ApiController` that all my API-controller then subclass instead of `ApplicationController`. `ApplicationController` could then inherit from `ActionController::Base` as usual.
(`ApplicationController` isn't an ancestor of any my API-controllers.)

ankane:
After thinking about this more, I think this will most likely compromise security. You can always override this in your app if needed with freedom patching.

PR #37, opened 2015-12-14
Bar chart option

petrvolny:
Since I don't believe that pie chart is the best visualization for most data cases I kinda missed an option to switch to bar chart so I implemented it.
Not sure if it is the best way how to implement it so I will be very happy to get any feedback.

Anyway, thanks for your work on this awesome project ;)

### DIFF
--- a/app/assets/javascripts/blazer/application.js
+++ b/app/assets/javascripts/blazer/application.js
@@ -1,5 +1,6 @@
 //= require ./jquery
 //= require ./jquery_ujs
+//= require ./jquery.query-object
 //= require ./list
 //= require ./stupidtable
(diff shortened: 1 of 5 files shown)


ankane:
Thanks for the pull request @petrvolny! I've been meaning to replace pie charts with column charts for a while. I like the clean approach you've taken and how you've kept the query parameters in sync.

In the interest of keeping the product simple (less options), I think we can just remove the pie chart.  I've done this on master and added support for multiple series. c51f7f091fa850c646e539d2286690f2abd9555c

petrvolny:
Thanks for your response and thumbs up for replacing pie charts with column charts ;)

PR #42, opened 2016-01-11
Prevent accidental backspace navigation on query edit page

righi:
When editing a query I often mistakenly think the text area has focus and hit backspace (delete). This navigates me back to the previous page and away from a very gorgeous and very unsaved query.

The code in this PR prevents backspaces (deletes) from doing that. (Only on the edit query page.)


ankane:
:100: I hate it when this happens.

PR #43, opened 2016-01-12
Preserve columns with the same name

righi:
Consider this query: 
```sql
select 
   city.id, city.name, country.id, country.name 
from 
   city 
   inner join country on city.country_id = country.id
```

Previously, Blazer would only display two columns: `id, name`

With this change, the same query will now return: `id_1, name_1, id_2,
name_2`

smart_columns and linked_columns will continue to work, as the original column name is preserved for their use.


ankane:
Nice, this has been on my list for a while.  Can we preserve the original column names in the results as SQL clients do (`id, name, id, name`)?

PR #44, opened 2016-01-23
Autolink URLs

righi:
Turns URLs in query results into links.

**Example SQL:**
```sql
SELECT 'https://github.com' AS url;
```

**Blazer Output:**
```
url
---
<a href="https://github.com" target="_blank">https://github.com</a>

### DIFF
--- a/app/helpers/blazer/base_helper.rb
+++ b/app/helpers/blazer/base_helper.rb
@@ -1,3 +1,5 @@
+require 'rails_autolink/helpers'
+
 module Blazer
   module BaseHelper
     def blazer_title(title = nil)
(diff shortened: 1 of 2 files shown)


ankane:
Thanks for the pull request.  This is a nice feature, but I'm afraid it'll slow down large responses significantly, so we could either:

1. Only do this for columns that end in `url`
2. Only do this when the number of rows is small

ankane:
Added here https://github.com/ankane/blazer/commit/d59da0ab687ff98824e23396e9170e6b1799898b for columns that are entirely urls so we don't need to introduce an additional dependency.

PR #45, opened 2016-01-24
Use limitcount for column chart

sarchak:
Use limit count for displaying distribution charts cc @righi @ankane


ankane:
Thanks Shrikar!  At some point, it becomes hard to read charts with many columns, so I set 20 as the max. Do you think this should be higher?

ankane:
Hey @sarchak, any thoughts on the comment above?

ankane:
Cleaning up issues

PR #46, opened 2016-02-21
Adds optional before_action configuration for blazer controller authentication/authorization

markrebec:
Older versions of devise + rails don't allow use of the `authenticate` method in your `config/routes.rb`, which means the only option for auth in blazer under that setup is http basic auth.

This adds support for an optional `before_action` configuration which allows specifying a method name as a symbol. That method will then be called as a `before_action` in the `Blazer::BaseController` so you can plug in your own auth flow (such as calling devise's `authenticate!`) or really do whatever you want.

### DIFF
--- a/app/controllers/blazer/base_controller.rb
+++ b/app/controllers/blazer/base_controller.rb
@@ -9,8 +9,8 @@ module Blazer
       http_basic_authenticate_with name: ENV["BLAZER_USERNAME"], password: ENV["BLAZER_PASSWORD"]
     end
 
-    if Blazer.auth_filter
-      before_action Blazer.auth_filter
+    if Blazer.before_filter
+      before_action Blazer.before_filter
     end
 
     layout "blazer/application"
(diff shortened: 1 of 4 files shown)


ankane:
Awesome, thanks Mark!  This is a nice feature.  What do you think about naming it `before_action`?

buren:
:+1: for this! I'm not using Devise and it would be awesome to be able to use the same authentication as for the rest of the app :)

markrebec:
I'm into `before_filter`, the more generic name is appropriate since technically you could do whatever you want with this hook. Done and pushed!

markrebec:
Ha, whoops, I've been jumping between rails 3/4 so much lately I totally mixed up `before_filter` and `before_action` in my update here :-P 

Andrew, lemme know if you feel strong enough about `before_action` vs. `before_filter` and I'll make one more quick commit with that name instead.

buren:
@markrebec I would argue that `before_action` is better :neckbeard: Mainly because rails moved from `before_filter` -> `before_action`, ultimately is @ankane choice ;)

markrebec:
@buren :+1: agreed, updated and pushed.

buren:
@markrebec :crown:

ankane:
Awesome, thanks Mark!

PR #47, opened 2016-02-25
fix statement issue (414 Request-URI Too Large) for long sql statement at Download button

dangnm:
SQL string in @statement can be very large to pass into a URL. Request URI is limited in about 8KB which depends on server configuration or browser. When statement size exceeds the limit, it will cause the error 414 Request-URI Too large.   So I think it should be passed as a body param.


ankane:
Awesome, thanks Dang! :+1:

PR #51, opened 2016-03-04
Fix tables

bernardolm:


bernardolm:
No formatting with delimiter columns named id or starts with id

PR #67, opened 2016-06-18
Add the ability to only query selected SQL

ilstar:
Many times I saw people comment/uncomment different queries in a same window for testing, this could make their (including me) life easier :)

What it does:
1. if user haven't selected any text, it works exactly as before
2. if user select one or two lines SQL, it only run that part

screenshot 2016-06-17 20 54 55


ankane:
Great, thanks Fred 👍

giedriusr:
Great feature 👍

PR #69, opened 2016-06-23
Render images for columns that end in 'image'

bonsaiben:
Renders image URLs in query results and links to the image.

Example SQL:

```
SELECT 'http://www.github.com/image.jpg' AS github_image;
```

Blazer Output:

```
github_image
---
<a href="http://www.github.com/image.jpg" target="_blank"><img src="http://www.github.com/image.jpg" width="200"></a>
```

image

### DIFF
--- a/app/helpers/blazer/base_helper.rb
+++ b/app/helpers/blazer/base_helper.rb
@@ -37,6 +37,10 @@ module Blazer
       ENV["MAPBOX_ACCESS_TOKEN"].present?
     end
 
+    def blazer_image?(key, value)
+      key.to_s.end_with?("image") && value.is_a?(String) && %w[png jpg jpeg].include?(value.split(".").last)
+    end
+
     JSON_ESCAPE = { '&' => '\u0026', '>' => '\u003e', '<' => '\u003c', "\u2028" => '\u2028', "\u2029" => '\u2029' }
     JSON_ESCAPE_REGEXP = /[\u2028\u2029&><]/u
 
--- a/app/helpers/blazer/base_helper.rb
+++ b/app/helpers/blazer/base_helper.rb
@@ -38,7 +38,7 @@ module Blazer
     end
 
     def blazer_image?(key, value)
-      key.to_s.end_with?("image") && value.is_a?(String) && %w[png jpg jpeg].include?(value.split(".").last)
+      key.to_s.end_with?("image") && value.is_a?(String) && %w[png jpg jpeg gif].include?(value.split(".").last.split("?").first)
     end
 
     JSON_ESCAPE = { '&' => '\u0026', '>' => '\u003e', '<' => '\u003c', "\u2028" => '\u2028', "\u2029" => '\u2029' }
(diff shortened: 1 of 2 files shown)


app/helpers/blazer/base_helper.rb
```diff
@@ -37,6 +37,10 @@ def blazer_maps?
       ENV["MAPBOX_ACCESS_TOKEN"].present?
     end
 
+    def blazer_image?(key, value)
+      key.to_s.end_with?("image") && value.is_a?(String) && %w[png jpg jpeg].include?(value.split(".").last)
```
shiftb: + gif please

shiftb: Also, would be good to look at extension between `.` and `?` so that urls like `http://.../image.gif?w=100&h=100` work

app/views/blazer/queries/run.html.erb
```diff
@@ -124,6 +124,10 @@
                           <div class="text-muted">empty string</div>
                         <% elsif @linked_columns[k] %>
                           <%= link_to blazer_format_value(k, v), @linked_columns[k].gsub("{value}", u(v.to_s)), target: "_blank" %>
+                        <% elsif blazer_image?(k, v) %>
+                          <%= link_to blazer_format_value(k, v), target: "_blank" do %>
+                            <%= image_tag blazer_format_value(k, v), width: '200px' %>
```
ankane: Thoughts on also displaying the url?

bonsaiben: Hmm, you could select for it like this:

```
SELECT 'http://www.github.com/image.jpg' AS github_image, 'http://www.github.com/image.jpg' AS github_image_url;
```

ankane:
We'll probably need to run this through Camo https://github.com/atmos/camo

bonsaiben:
@kwent not much cleaner IMO, and regexp is 50% slower (2.2.4p230)

```
> n = 10000000
> Benchmark.bm do |x|
>   x.report { n.times do; 'http://ia.media-imdb.com/images/M/MV5BMTQ2Mjc1MDQwMl5BMl5BanBnXkFtZTcwNzUyOTUyMg@@._V1_UY1200_CR90,0,630,1200_AL_.jpg?what=what'.split(".").last.split("?").first; end }
>   x.report { n.times do; 'http://ia.media-imdb.com/images/M/MV5BMTQ2Mjc1MDQwMl5BMl5BanBnXkFtZTcwNzUyOTUyMg@@._V1_UY1200_CR90,0,630,1200_AL_.jpg?what=what'.match(/[^\s]+(\.(jpg|png|gif|bmp))/)[2]; end }
> end
       user     system      total        real
  20.110000   0.090000  20.200000 ( 20.232612)
  30.130000   0.220000  30.350000 ( 30.389175)
```

ankane:
I'm a bit worried about the security repercussions of this (including possibly "malicious" images), but couldn't find much on the topic.  Here's one discussion: http://security.stackexchange.com/questions/36447/img-tag-vulnerability

What do you think of a whitelist to limit the possible attack vector?

app/views/blazer/queries/run.html.erb
ankane: Any reason to use `blazer_format_value` instead of just `v`?

ankane: Also, it may be good to use `referrerpolicy="no-referrer"` for privacy, although browser support is limited.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img

ankane:
Added behind `images` setting. https://github.com/ankane/blazer/commit/6dcc5cf347523976216e76786a2c71d445fd0f72

PR #71, opened 2016-07-07
Fixbug: it shows incorrect error line num when we only query selected SQL

ilstar:
How to reproduce:

1. write 2 queries and make sure the 2nd query has a syntax error
1. select the 2nd query and run it
1. then you will notice the error line is incorrect, we should add a "padding" on it

Here is the before vs after, sorry for the issue I involved last time.
pjimage


ankane:
Thanks Fred! 👍

PR #72, opened 2016-07-14
Add audits page as a "hidden feature"

ilstar:


ankane:
Hey @ilstar, is there any major advantage to having a separate page vs writing a Blazer query for this?

ilstar:
Ah, I didn't think about we could use blazer to do that, such a dumb PR, I will close it.

--
Fred Liang
Sent with Sparrow

ankane:
No worries 👍

PR #74, opened 2016-07-28
Fix to no formatting with delimiter columns named id or starts with id

bernardolm:
This helps when you works with a legacy databases


ankane:
Thanks @bernardolm. No need to check if `key == "id"` as both start_with and end_with cover that case.

bernardolm:
It's true, @ankane. Fixed.

renatoelias:
👍

ankane:
Looks good, thanks @bernardolm 👍

bernardolm:
;D

PR #75, opened 2016-07-30
Remove `before_action :verify_request_size` from BaseController

jpteti:
Best I can tell, this line was carried over from Ahoy by accident in commit 4e973845656a547c8330f336772f6be326e3a971. There is no `verify_request_size` method in Blazer, so this was raising an exception.


ankane:
Whoops, poor mistake. Will push out a new version. Thanks for fixing @jpteti 👍

PR #77, opened 2016-08-02
Add lat/lng as mappable column keys

coreymartella:


ankane:
Looks good, thanks @coreymartella 👍

PR #20, opened 2015-08-26
Support smart_variables as a Hash or Array

starchow:
Currently, the smart_variables must be an SQL query. Sometimes I know the exactly values and I want write it as a Hash or Array.

I also updated README to easier understand how to use linked_columns, smart_columns, smart_variables.

Please check out my pull request and merge it if it good enough. 
Thanks!

PR #78, opened 2016-08-02
Day of week names

coreymartella:
Not sure of your opinion here. We have a lot of weekday stats but the `gd_day_of_week` sql function returns an int so that needs to be converted in order to end up in charts and easily readable datatable. Figured the hash style `smart_columns` mapping could be useful for other purposes/non-db enums while falling back to the I18n implementation for `day_of_week` named columns.

If you'd like only go with one of the approaches or perhaps not even I18n just `Date::DAY_NAMES` I'm open to suggestions

### DIFF
--- a/README.md
+++ b/README.md
@@ -220,6 +220,20 @@ smart_columns:
   city_id: "SELECT id, name FROM cities WHERE id IN {value}"
 ```
 
+You can also put a Hash into the config for more static data:
+
+```yml
+smart_columns:
+  day_of_week:
+    0: Sun
+    1: Mon
+    2: Tue
+    3: Wed
+    4: Thu
+    5: Fri
+    6: Sat
+```
+
 ### Caching
 
 Blazer can automatically cache results to improve speed. It can cache slow queries:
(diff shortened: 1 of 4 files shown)


ankane:
I like the idea of allowing a hash for smart columns (this was proposed in #20 as well). I'd say we go with that. I don't see a huge benefit of making a special case for `day_of_week`.

coreymartella:
Only benefit I see is I18n but nothing else in the gem is localized so thats a stretch. All users get it for free instead of having to define the static mapping for each install would be the other but I guess if day of week is important to them the definition in the config is trivial.

I'll back out the other changes and re-push. I also have a non-PRed change for date & time formatting and right aligned numerics using similar logic in my [fork](https://github.com/coreymartella/blazer/compare/master...coreymartella:customizations) but happy to discuss pulling it in if interested. Thanks for the great work, this gem (and your others) are super useful.

ankane:
Thanks @coreymartella. For arrays, I'd expect it to use the value for the key as well rather than the index.

```yml
smart_columns:
  period: ["day", "week", "month"]
```

ankane:
Hey @coreymartella, any progress?

PR #79, opened 2016-08-04
Allow downloading csv via GET request

wheeyls:
When using our BI system with R, it'd be really nice if we could expose the CSV downloads to a GET request, so all we have to do is call `read.csv('http://user:pass@example.com/queries/run.csv?...')` to get at the data.

This is obviously a little awkward the way that this is set up, so I'm opening the PR to start the conversation. I don't want the raw SQL query to be in my link, I want the query to be constructed from parameters the same way it is on `queries#show`. In fact, would it make sense for `queries#show.csv` to return the CSV results of the run?

### DIFF
--- a/app/views/blazer/queries/show.html.erb
+++ b/app/views/blazer/queries/show.html.erb
@@ -14,7 +14,7 @@
         <%= link_to "Fork", new_query_path(variable_params.merge(fork_query_id: @query.id, data_source: @query.data_source, name: @query.name)), class: "btn btn-info" %>
 
         <% if !@error && @success %>
-          <%= button_to "Download", run_queries_path(query_id: @query.id, format: "csv"), params: {statement: @statement}, class: "btn btn-primary" %>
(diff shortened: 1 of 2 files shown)


ankane:
Hey @wheeyls, thanks for getting this conversation started.  On a high level, I like the idea of Blazer and R being friends.  However, when I think about the use cases, I struggle to see a reason to use Blazer vs accessing the database directly.

About the implementation: I chose to implement the query endpoints as POST rather than the RESTfully correct GET to reduce the possibility of a CSRF or denial of service attack.

wheeyls:
It's a good point that I could just hit my database directly from R. I guess the reason I haven't been doing that is that my R guy doesn't know or really care about the structure of our database. It's nice to be able to just give him a time series and let him get to work. Of course I can just have him download the CSV and that works fine.

But when I used this tool though I immediately felt myself reaching for the output as a sharable URL. R was one use case, the other immediate one is to provide data for our existing metrics dashboards - things like Geckoboard.

If you think that's stretching this tool outside of what it should be good at, I think that's probably fair.

ankane:
@wheeyls Yeah, I'd probably say it's out of scope for now.

PR #82, opened 2016-08-18
add JSON output to query#show

blairanderson:
- converts `download` button to `csv`
- adds `json` button.

screen shot 2016-08-17 at 11 38 39 pm
screen shot 2016-08-17 at 11 51 25 pm

### DIFF
--- a/app/controllers/blazer/queries_controller.rb
+++ b/app/controllers/blazer/queries_controller.rb
@@ -232,6 +232,9 @@ module Blazer
         format.csv do
           send_data csv_data(@columns, @rows, @data_source), type: "text/csv; charset=utf-8; header=present", disposition: "attachment; filename=\"#{@query.try(:name).try(:parameterize).presence || 'query'}.csv\""
         end
+        format.json do
+          render json: json_data(@columns, @rows, @data_source)
+        end
       end
     end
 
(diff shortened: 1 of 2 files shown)


ankane:
Hey Blaire, thanks for the PR. What are the main use cases for exporting JSON? In the interest of keeping the UI simple, I'd prefer not add additional buttons unless they'll be commonly used.

On a separate note, if there are multiple columns with the same name in the results, it looks like the last column will be the only one to show up in the JSON. An alternative approach would be to format it like:

```json
{
  "columns": ["column_a", "column_b"],
  "rows": [
    ["value1", "value2"],
    ["value3", "value4"]
  ]
}
```

But I'm not sure it feels as nice.

blairanderson:
Hey kane, 
I think blazer is super cool because it helps people make data available. Adding more formats seems like a way to continue that.

Main use-case: Blazer as an api for demo's and such.

cycomachead:
Yeah, I could totally see myself using a JSON output, especially as a way of making a "fake" API for some projects.

In terms of UI, you could make "Download" be a segmented button, with a dropdown arrow (V) that allows selecting JSON instead of CSV. I don't know if there would ever be other formats, but that would easily extend to at least a few different types.

holden:
I think this would be useful and very cool.

ankane:
Going to close. I'm personally not a fan of adding more download formats, as pretty much all languages have CSV support.

PR #92, opened 2016-10-28
suppress deprecation warnings when creating tables

ts-3156:
Thank you for super useful gem!

This pull request fixes these warnings.

```
-- create_table(:blazer_queries)
DEPRECATION WARNING: `#timestamps` was called without specifying an option for `null`. In Rails 5, this behavior will change to `null: false`. You should manually specify `null: true` to prevent the behavior of your existing migrations from changing. (called from block in change at [rails root]/db/migrate/20161021113742_install_blazer.rb:9)
   -> 0.0197s
-- create_table(:blazer_audits)
   -> 0.0105s
-- create_table(:blazer_dashboards)
DEPRECATION WARNING: `#timestamps` was called without specifying an option for `null`. In Rails 5, this behavior will change to `null: false`. You should manually specify `null: true` to prevent the behavior of your existing migrations from changing. (called from block in change at [rails root]/db/migrate/20161021113742_install_blazer.rb:23)
   -> 0.0125s
-- create_table(:blazer_dashboard_queries)
DEPRECATION WARNING: `#timestamps` was called without specifying an option for `null`. In Rails 5, this behavior will change to `null: false`. You should manually specify `null: true` to prevent the behavior of your existing migrations from changing. (called from block in change at [rails root]/db/migrate/20161021113742_install_blazer.rb:30)
   -> 0.0110s
-- create_table(:blazer_checks)
DEPRECATION WARNING: `#timestamps` was called without specifying an option for `null`. In Rails 5, this behavior will change to `null: false`. You should manually specify `null: true` to prevent the behavior of your existing migrations from changing. (called from block in change at [rails root]/db/migrate/20161021113742_install_blazer.rb:42)
   -> 0.0132s
```


ankane:
Thanks @ts-3156 👍 

I think we should go with `null: false`, since that's the default as of Rails 5. Also, it shouldn't be needed for the `last_run_at` column.

ts-3156:
Thank you for replying. I changed it to `null: false`. This migration works for me on localhost.

ankane:
Looks good, thanks @ts-3156! 💯

PR #95, opened 2016-11-01
Add date_range option

ts-3156:
Hello,

This PR adds date_range option to specify default date_range value.
It's useful when the table size is too big.

Thanks.

----

This PR is not related, but, I'm sorry and thank you -> 36f43257917e30c00de7f8de9995fc225eb63ec5 .


ankane:
Hey, thanks for another PR 👍 

This doesn't seem like a common use case, and I'm not a huge fan of the current date picker, so I'd prefer not to add options around it.

Also, try to keep changesets as small as possible. It makes them much easier to review.

ts-3156:
Thank you for replying. I understand the reason. I am sorry for the inconvenience, but please close this PR.

>> Also, try to keep changesets as small as possible. It makes them much easier to review.

Thank you for the advice. I will take care in future.

ankane:
Hey @ts-3156, don't worry, not an inconvenience - just giving feedback for future PRs (for this and other open source projects).

PR #105, opened 2016-11-16
Add prefix to query name when it forked

ts-3156:
Hello,

When I fork a query, I always confuse origin query with forked query. To prevent this mistake, I tried to add prefix `Forked` to query name.

Please let me know your thoughts.

Thanks!

### DIFF
--- a/app/controllers/blazer/queries_controller.rb
+++ b/app/controllers/blazer/queries_controller.rb
@@ -39,7 +39,9 @@ module Blazer
         name: params[:name]
       )
       if params[:fork_query_id]
-        @query.statement ||= Blazer::Query.find(params[:fork_query_id]).try(:statement)
+        query = Blazer::Query.find_by(id: params[:fork_query_id])
+        @query.statement ||= query.try(:statement)
+        @query.name = "Forked: #{query.try(:name)}"
       end
     end
 


ankane:
Hey @ts-3156, thanks for the PR 👍 

Blazer used to do this (with "Fork of ___"), but you'd always end up deleting it out of the name. Maybe we could add a different indicator. "Fork" next to the label is one idea.

ts-3156:
Hello, Thank you for your reply!

> Blazer used to do this (with "Fork of ___"), but you'd always end up deleting it out of the name.

Thank you for this information. Come to think of it, maybe you are right. I suppose the points I think it's confusing consist of the following factors.

1. The fork buttons have different roles

The fork buttons are placed on `show` page and `edit` page. When I click the former, I go to `new` page. That makes sense. When I click the latter, new query is saved. It's not intuitive because I expect to got to `new` page. In order to solve this unexpected behavior, I suggest removing the fork button on `edit` page.

2. I can save duplicate query name

When I fork a query, I accidentally save duplicate query name. I'm happy if a warning dialog is shown after I click save button for duplicate query name and only when I click ok, the query is saved.

Please let me know your thoughts.

Thanks!

ankane:
Hey @ts-3156, sorry the discussion on this stalled.

For 1) you can choose to fork before or after editing. I think both options are valuable (I used them all the time), so don't think we should change that.

For 2) I think that could make sense, but am worried about what adding that query will do for existing installation that won't have an index on name.

Overall, I'm not sure this is a big enough issue to address right now.

ts-3156:
Thank you for polite response!

PR #114, opened 2017-01-25
Export smart column map value in CSV

dwhenry:
This is something that we need for the exported data to have value. 

I feel like it could potential be better to have this be configurable functionality, however I am not sure the best place to put the configuration (thinking either the query object or potentially globally in the blazer.yml file).


ankane:
Hey @dwhenry, thanks for the PR 👍  I think global opt-in makes sense for this. Also, how about a `_label` suffix instead of `_map`?

dwhenry:
Unable to push new changes to Repository.

PR #116, opened 2017-02-02
Add mssql server adapter

### LINKED ISSUE #115: Preview tables for MS SQL Server doesn't work
There are some hard coded queries in the `SqlAdapter` that are not compatible with SQL Server.

Functions like listing the schema and table do not work.

ttilberg:
This adds compatibility for MS SQL Server : fixes #115 

This PR includes adding Minitest according to the [Rails Guides](http://edgeguides.rubyonrails.org/engines.html#testing-an-engine), as currently there is not a testing framework in place. It uses the dummy app at `blazer/test/dummy` as "the app that mounts this engine." 

By adding a testing framework, the gemspec now has some additional development dependencies related to DBs. If you would like a separate PR without the testing framework, please let me know so I can modify.

I've used this in a local app against SQL Server and it looks like it worked great. 

The tests require an instance of SQL Server.

Testing with the SQL Server adapter is not on by default, you must set `WITH_SQLSERVER` to anything. This is due to the fact that most times tests will be run in the future, there is no SQL Server client to connect to.

```bash
$ BLAZER_DATABASE_URL='sqlserver://rails:railspassword@localhost/master' WITH_SQLSERVER=true bundle exec rake test`
```

If you do not have SQL Server installed locally, but still want to hack on this, you can use Docker:

```bash
docker run -e ACCEPT_EULA=Y -e SA_PASSWORD=GreatPassword! -p 1433:1433 --rm microsoft/mssql-server-linux

# Open a new terminal, as this hogs the front end and we want to --rm it, so no -d
BLAZER_DATABASE_URL=sqlserver://sa:GreatPassword\!@localhost/master WITH_SQLSERVER=true bundle exec rake test
# The default container has a strong password requirement, so be sure to escape anything bash doesn't like.
```

This is my first code based public PR, so I'd appreciate whatever feedback you have to offer.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,9 @@
 *.o
 *.a
 mkmf.log
+log/*.log
+test/dummy/db/*.sqlite3
+test/dummy/db/*.sqlite3-journal
+test/dummy/log/*.log
+test/dummy/tmp/
+test/dummy/.sass-cache
--- a/Rakefile
+++ b/Rakefile
@@ -1,1 +1,34 @@
-require "bundler/gem_tasks"
+begin
+  require 'bundler/setup'
+rescue LoadError
+  puts 'You must `gem install bundler` and `bundle install` to run rake tasks'
+end
+
+require 'rdoc/task'
+
+RDoc::Task.new(:rdoc) do |rdoc|
+  rdoc.rdoc_dir = 'rdoc'
+  rdoc.title    = 'Blorgh'
+  rdoc.options << '--line-numbers'
+  rdoc.rdoc_files.include('README.rdoc')
+  rdoc.rdoc_files.include('lib/**/*.rb')
+end
+
+
+
+
+
+
+Bundler::GemHelper.install_tasks
+
+require 'rake/testtask'
+
+Rake::TestTask.new(:test) do |t|
+  t.libs << 'lib'
+  t.libs << 'test'
+  t.pattern = 'test/**/*_test.rb'
+  t.verbose = false
+end
+
+
+task default: :test
--- a/blazer.gemspec
+++ b/blazer.gemspec
@@ -12,9 +12,9 @@ Gem::Specification.new do |spec|
   spec.homepage      = "https://github.com/ankane/blazer"
   spec.license       = "MIT"
 
-  spec.files         = `git ls-files -z`.split("\x0")
+  spec.files         = Dir["{app,config,db,lib}/**/*", "LICENSE.txt", "Rakefile", "README.md"]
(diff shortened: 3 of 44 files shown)


ankane:
Hey @ttilberg, thanks for jumping on that so quickly :) Great to know there's a SQL docker image to test this on. Here are a few best practices for PRs:

1. Keep the changeset as small as possible. Leave out non-critical dependencies, like Pry, and do a separate PR for each feature, like adding a test framework.

2. Try to follow the same style. Rather than creating a new adapter that extends `SqlAdapter`, add the logic to `SqlAdapter` (see how Redshift is handled).

A few things specific to this PR:

- We should test cancel before adding it. I typically use sleep functions to do this. It looks like SQL Server has `WAITFOR DELAY '00:01'`.
- Is there a reason for overriding `preview_statement`?
- I'd rather not add a test framework at this time.

Looking forward to getting this merged on the above is addressed!

ttilberg:
Okay, I'll make these changes tomorrow morning and resubmit. For now, a few quick answers to your questions: 

- Happy to remove he test framework- it was mostly there to prove it works as expected both to me and you. 


- I opted for inheritance over adding more if statements as I felt it was starting to become a little overwhelming with the many if/elses and such. Inheriting from the generic sql adapter seemed more clean to me. Happy to oblige though. 


- I'll see about cancelling a sleeping query in the morning: makes sense and should be easy to track down. Thanks for the suggestion. 

- Regarding the preview statement: `limit 10` isn't valid in MS SQL, so the preview statement uses the expected syntax: `top 10`

ankane:
Sounds good, I also updated the SQL Server instructions in the readme to make them more clear.

ttilberg:
Say, I was implementing `raise NotImplementedError` for `#cost` and `#explain` because I think I was interpreting a different part of the code incorrectly:

```ruby
# lib/blazer/run_statement.rb
if !result.cached? && duration >= 10
  audit.cost = data_source.cost(statement) if audit.respond_to?(:cost=)
end
```

This appears to be the only code that references these methods and initially I was reading it as `respond_to?(:cost)` not `(cost=)`. My original intent was to not return true for the condition.

After looking again, should this instead just return `nil` if I'm unable to get the stats? 

This evening I thought of something that might allow this to be implemented after all, but I'd like to know how you'd like it handled if not.

ankane:
You shouldn't need to do anything with cost. The existing logic will return `nil` for adapters other than Postgres or Redshift. This feature isn't used by default (no cost column on `blazer_audits`) - it was just something I was trying out.

ankane:
Without query cancellation (which we could do in a separate PR), I think the changeset can be under 10 lines - just additional conditions in `preview_statement` and `schemas`.

PR #119, opened 2017-02-03
Add syntax for MS SQL Server

ttilberg:
This PR replaces #116 and closes issue #115

This addition is in the style requested and provides the following:

- Can now view DB schemas from app, and preview tables
  - Adds some MS SQL Server specific syntax
  - Sets default schema to `dbo`
- Adds MS SQL Server version of `#explain` and `#cost`

It does not yet implement `#cancel`-- perhaps for a future PR.

`#explain` returns a hash of values related to query performance, is this okay, or do you need it as a string? I didn't see `#explain` being accessed publicly anywhere.

The following tests passed:

```ruby
  class SqlServerAdapterTest < ActiveSupport::TestCase
    def setup
      @adapter = Blazer::Adapters::SqlAdapter.new(Blazer.data_sources["mssqlserver"])
    end

    test 'setup with a datasource' do
      assert_kind_of Blazer::Adapters::SqlAdapter, @adapter
    end

    test '#tables' do
      tables = @adapter.tables
      assert tables.count > 0, "Didn't find any tables. Do you have a SQL Server configured? If yes, this fails."
      assert_kind_of String, tables.first
    end

    test '#schema' do
      a_schema = @adapter.schema.first
      assert_not_nil a_schema[:table], "Didn't seem to find an expected entry in the schema list. Do you have a SQL Server configured? If yes, this fails."
    end

    test '#preview_statement' do
      assert @adapter.preview_statement =~ /SELECT TOP \(?\d+\)?/i
    end

    test '#cost(statement)' do
      cost = @adapter.cost 'SELECT * FROM information_schema.tables'
      assert_kind_of String, cost
      assert cost =~ /[0-9.]+/, "Cost was not a number"

      non_query_cost = @adapter.cost 'SELECT * FROM information_schema.tables'
      assert non_query_cost == "", "Cost should be empty."
    end

    test '#explain(statement)' do
      explain = @adapter.explain 'SELECT * FROM information_schema.tables'
      assert explain.include?("TotalSubtreeCost"), "SHOWPLAN_ALL is not on."
 
      # Now make sure it successfully disables SHOWPLAN_ALL for future queries
      after_explain_query = @adapter.run_statement('SELECT 1', 'no comment').each.first
      refute after_explain_query.include?("TotalSubtreeCost"), "SHOWPLAN_ALL is still active when it should not be."
    end

  end
```

### DIFF
--- a/lib/blazer/adapters/sql_adapter.rb
+++ b/lib/blazer/adapters/sql_adapter.rb
@@ -53,6 +53,8 @@ module Blazer
       def preview_statement
         if postgresql?
           "SELECT * FROM \"{table}\" LIMIT 10"
+        elsif sqlserver?
+          "SELECT TOP (10) * FROM {table}"
         else
           "SELECT * FROM {table} LIMIT 10"
         end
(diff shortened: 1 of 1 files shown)


ankane:
Looks much cleaner, thanks @ttilberg 💯

PR #124, opened 2017-02-10
Set schema search_path when schemas specified

bardsley:
Adding functionality to make schemas: [‘example’,’other’] work by default and make non standard schema names work as well

Whenever a blazer.yml has schemas for a datasource set the search path
for this connection to enable the access of this table with standard
SQL statements.

PostgreSQL specific

### DIFF
--- a/lib/blazer/adapters/sql_adapter.rb
+++ b/lib/blazer/adapters/sql_adapter.rb
@@ -13,6 +13,14 @@ module Blazer
             end
             establish_connection(data_source.settings["url"]) if data_source.settings["url"]
           end
+
+        setup_schemas
+      end
+
+      def setup_schemas
+        return if !postgresql? || schemas.empty?
+        schema_sql = 'SET search_path TO ' + schemas.join(',')
+        execute(schema_sql)
       end
 
       def run_statement(statement, comment)


bardsley:
Also have a pull request for data to test this in blazer-dev

ankane:
Hey @bardsley, thanks for the PR. I don't think Blazer should modify the `search_path`. You should be able to do that directly on the DB user if needed.

bardsley:
Hi @ankane, thanks for the feedback. I can see your point. The reason we put it in was after settings schemas in the blazer.yml you get get listings for tables that don't work when you select them by default. 

Another option would be to return table names in the form "schema"."table_name" whenever they aren't in the default schema. What do you think about that as an option?

ankane:
We could add the schema name if it's not the in `search_path` (I wouldn't want to just add it for all non-default schemas), but it feels like unnecessary complication if users can update their `search_path` independently.

ankane:
Cleaning up PRs

ACPK:
@ankane - Was a solution ever proposed to handle multiple schemas? For example, you might want to search for users in a multi-tenant app that uses a different schema for every multi-site.

ankane:
@ACPK Overall, Blazer works great with multiple schemas - it's just the table preview that's a bit off. I just pushed an `improved_schemas` branch that addresses it.

ACPK:
@ankane - Thanks! However is there a way to search all schemas? We use the apartment gem which created a new schema for every client. As such, we have over 150 schemas.

PR #125, opened 2017-02-15
Feature - query download adapters

JoWilfrid:
While downloading queries results, we can only have it in CSV and that's quite sad because it's not enhanceable from outside the gem.

I've done an adapter system that allows applications using `blazer` to define their own adapters to download queries in multiple formats (e.g: json, pdf, etc.).
Adapters classes are explicitly given while configuring Blazer, in `blazer.yml`.

2 "base" adapters are present in the app: `RenderAdapter` and `SendDataAdapter`.
Basically, every adapter that wants to render something inherits from `RenderAdapter` and every one that wants to make user download something inherits from `SendDataAdapter`.

I've also added some tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -712,6 +712,12 @@ audit: true
 # from_email: blazer@example.org
 ```
 
+## Tests
+
+```
+bundle exec rake test
+```
+
 ## TODO
 
 - advanced permissions
--- a/Rakefile
+++ b/Rakefile
@@ -1,1 +1,9 @@
 require "bundler/gem_tasks"
+require "rake/testtask"
+
+Rake::TestTask.new do |t|
+  t.name = :test
+  t.libs << "test"
+  t.libs << "app/adapters"
+  t.test_files = FileList['test/**/*_test.rb']
+end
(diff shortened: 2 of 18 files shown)


ankane:
Hey @JoWilfrid, thanks for another PR 👍  I'm personally not a fan of adding more download formats, as pretty much all languages have CSV support.

cycomachead:
> I'm personally not a fan of adding more download formats, as pretty much all languages have CSV support.

I see the appeal in only maintaining one output format (and for keeping the UI simple). What appeals to me about this method is that those of us who want to could write our own adapters, from really simple JSON ones to more complicated and specialized PDF renders. 

Your call, of course, but I would find an API like this useful and would hopefully get a chance to share any adapters I write.

JoWilfrid:
I did this because I needed to have the ability to download queries as JSON files.
Sure, I could have write a couple of ruby lines to convert CSV into JSON but this approach was supposing that me or another tech person was downloading the file to convert it. The fact is I need it simple for non-tech people.

As I think anyone could have this need one day, I opened this PR :)

ankane:
Hey @JoWilfrid, sorry the discussion stalled. I'd still prefer to err on the side of keeping things simpler, so going to close. Thanks again for putting this together 👍

PR #126, opened 2017-02-15
For existing query, add ability to run it without passing statement i…

JoWilfrid:
For an existing query, the statement is passed through parameters.
This commit adds the ability to get it (the statement) from the query if there is a query.

### DIFF
--- a/app/controllers/blazer/queries_controller.rb
+++ b/app/controllers/blazer/queries_controller.rb
@@ -74,12 +74,13 @@ module Blazer
     end
 
     def run
+      @query = Query.find_by(id: params[:query_id]) if params[:query_id]
       @statement = params[:statement]
+      @statement ||= @query.statement.dup if @query && @query.statement
       data_source = params[:data_source]
       process_vars(@statement, data_source)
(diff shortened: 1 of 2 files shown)


JoWilfrid:
As the statement is not required anymore in the parameters, the risk to have an URL > 255 chars is low, so I allowed the method to be called via HTTP GET request.

ankane:
Hey @JoWilfrid, thanks for the PR. Unfortunately, this opens Blazer up to CSRF. See http://guides.rubyonrails.org/security.html#cross-site-request-forgery-csrf

What did you want this for?

JoWilfrid:
Oh, you're right! As the project I'm using for testing my modifications in Blazer's project is using `rack-jwt` to prevent strangers to acces it, I didn't realize that. Sorry about it!

> What did you want this for?

For the fact of requesting the `run` method without passing statement, I wanted this to be able to request a query result without having to know its statement.

By accessing via GET, I just thought that you did it via POST because of the potential very large size of the statement param which would not be supported by all browsers if too big.
For this point it looks like I was wrong.

PR #127, opened 2017-02-19
Fix dashboard queries form when wrapped under root key

### LINKED ISSUE #108: Dashboard edit view broken
I'm seeing 👇,  when editing a dashboard. All the names are blank in the list and if I try to submit the form `POST /blazer/dashboards/` I get `404 Not found`.

screen shot 2016-12-13 at 01 40 41

---

I quickly dug into the code and `var dashboardQueries` thats generated by `blazer_js_var`, seems correct. However after Vue has been initialised, `dashboardQueries` is now a Vue object and I've never used `Vue` before so didn't dig any further.

Blazer v1.7.5
I've tested in Chrome 55, Firefox 46 and Safari 9, so it doesn't look like a browser bug.

Thanks for an awesome gem 🙇

buren:
I have `ActiveRecord.Base.include_root_in_json = true`, so I guess that could be it.

---
Closes #108.

### DIFF
--- a/app/helpers/blazer/base_helper.rb
+++ b/app/helpers/blazer/base_helper.rb
@@ -33,7 +33,7 @@ module Blazer
     end
 
     def blazer_js_var(name, value)
-      "var #{name} = #{blazer_json_escape(value.to_json)}".html_safe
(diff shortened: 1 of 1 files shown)


ankane:
Great, thanks @buren!

PR #130, opened 2017-02-28
Export smart column map

dwhenry:
@ankane sorry for creating a new pull request, I have changed jobs and no longer have write access to the last PR. 

#114 

I have updated this based on your feedback:

* Added global config variable `output_smart_columns_in_csv`
* renamed columns that don't end in "_id" with "_label"

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,6 +235,13 @@ smart_columns:
   status: {0: "Active", 1: "Archived"}
 ```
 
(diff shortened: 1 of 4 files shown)


dwhenry:
@ankane any hope of getting merged or should i just close the PR??

dwhenry:
@ankane I think this would add value addition to your project but don't want to leave it as an open PR anymore.

Let me know if you agree and I will reopen the PR.

PR #131, opened 2017-03-02
Deterministic sort for floats with empty values in Firefox

jstastny:
The sorting of stupidtable was not deterministic for columns with `float` and `NULL` values in Firefox.

*Before:*
image

*After:*
image


ankane:
Great, thanks @jstastny 👍

PR #133, opened 2017-03-17
Update installation instruction on README

davidslee101:
I added a line to run 'bundle install' before running 'rails g blazer:install'

I typically copy and paste installation instructions line by line, and I got an error the first time I tried to install before I realized that I didn't run 'bundle install'

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ gem 'blazer'
 Run:
 
 ```sh
+rails bundle install
 rails g blazer:install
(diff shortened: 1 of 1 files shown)


ankane:
Hey @davidslee101, thanks for the PR 👍  Think it should just be `bundle install` without the `rails`.

davidslee101:
ah you're right. I'll close this down and resubmit!

Issue #66, opened 2016-06-16
Blazer app colliding with my main ApplicationController and app's routes


aguynamedben:
Hi, I'm trying to install blazer to try it out. I'm running into a few weird issues where Blazer is colliding with my main app.

My core question is: Is Blazer supposed to be mounted as an entire separate app, similar to Sidekiq's web interface? Or is Blazer supposed to inherit code (controller, routes) from my main app?

Environment:
- Ruby 2.3.1
- Rails 4.2.6
- Blazer 1.4.0

I followed the README very directly...
- Default blazer.yml generated by `rails g blazer:install`
- `mount Blazer::Engine, at: "blazer"` in config/routes.rb

The first problem I run into is this error when loading /blazer:

```
You must activate the Authlogic::Session::Base.controller with a
controller object before creating objects
```

My main app uses the authlogic gem. The traceback for this error leads to a method in app/controllers/application_controller.rb (my main application controller). `Blazer::BaseController` subclasses `ApplicationController`, and it's using my main app's `ApplicationController`. Is this intentional?

When I comment out the offending authentication function in `ApplicationController`, the /blazer path loads, but all URL generation on the page fails and none of the links work (i.e. can't a new dashboard). The log says:

```
DEPRECATION WARNING: You are trying to generate the URL for a named route called "new_query"
but no such route was found. In the future, this will result in an `ActionController::UrlGenerationError`
exception.` `rake routes` shows the routes like this:
```

`rake routes` shows:

```
(my apps routes, where there is no new_query route)

Routes for Blazer::Engine:
      run_queries POST   /queries/run(.:format)            blazer/queries#run
    refresh_query POST   /queries/:id/refresh(.:format)    blazer/queries#refresh
   tables_queries GET    /queries/tables(.:format)         blazer/queries#tables
   new_query GET    /queries/new(.:format)            blazer/queries#new
...
```

Somehow, it seems like Blazer is again, when looking for routes, looking in the routes for my app, not the Blazer Engine routes. From reading about Rails Engines, it seems like the intention is for none of Blazer's code to use my ApplicationController or routes, but that's not happening.

It's too bad, I wish I could get this working to test. I'm happy to do some extra work around here if you guide me on where to look. Thanks for the time you spend working on this gem, it looks awesome, I hope I can get it working somehow.

ankane:
[...]

PR #140, opened 2017-03-28
Preserve Authlogic before_actions

cannikin:
Deals with issue #66 — a simple implementation that works great.

I thought about exposing the list of filters to preserve as a config variable, but if this is the first time this issue has come up (removing too many filters) it's probably not worth complicating the config logic and documentation yet.

Let me know what you think!

### DIFF
--- a/app/controllers/blazer/base_controller.rb
+++ b/app/controllers/blazer/base_controller.rb
@@ -1,7 +1,13 @@
 module Blazer
   class BaseController < ApplicationController
-    # skip all filters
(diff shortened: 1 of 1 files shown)


ankane:
Thanks @cannikin 👍  I think this can be a bit simpler by adding `- [:activate_authlogic]` to the current line.

cannikin:
Even simpler!

ankane:
Thanks @cannikin.

PR #141, opened 2017-04-03
Use multiple values in smart variables

pedrocarmona:
Hi @ankane,

This PR allows the user to select multiple values from a smart_variable.

Its possible to use smart_variables in queries with IN.
```
SELECT ...
FROM ...
WHERE user_id IN({user_ids})
```

### DIFF
--- a/app/assets/javascripts/blazer/selectize.js
+++ b/app/assets/javascripts/blazer/selectize.js
@@ -2707,7 +2707,7 @@
           label = self.options[self.items[i]][self.settings.labelField] || '';
           options.push('<option value="' + escape_html(self.items[i]) + '" selected="selected">' + escape_html(label) + '</option>');
         }
-        if (!options.length && !this.$input.attr('multiple')) {
+        if (!options.length) {
           options.push('<option value="" selected="selected"></option>');
         }
         self.$input.html(options.join(''));
--- a/app/controllers/blazer/base_controller.rb
+++ b/app/controllers/blazer/base_controller.rb
@@ -33,27 +33,35 @@ module Blazer
 
         if @success
           @bind_vars.each do |var|
-            value = params[var].presence
-            if value
-              if ["start_time", "end_time"].include?(var)
-                value = value.to_s.gsub(" ", "+") # fix for Quip bug
-              end
+            values = Array(params[var])
+            values.each_with_index do |value_str|
+              value = value_str.presence
+              if value
+                if ["start_time", "end_time"].include?(var)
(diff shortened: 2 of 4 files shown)


app/controllers/blazer/base_controller.rb
```diff
@@ -33,27 +33,35 @@ def process_vars(statement, data_source)
...
+                  value = value.to_i
+                elsif value =~ /\A\d+\.\d+\z/
+                  value = value.to_f
+                end
+              end
+              value = ActiveRecord::Base.connection.quote(value)
+              if values.last == value_str
+                statement.gsub!("{#{var}}", value)
+              else
+                statement.gsub!("{#{var}}", "#{value}, {#{var}}")
```
pedrocarmona: Should this be also activated only for plural variables?

app/assets/javascripts/blazer/selectize.js
```diff
@@ -2707,7 +2707,7 @@
           label = self.options[self.items[i]][self.settings.labelField] || '';
           options.push('<option value="' + escape_html(self.items[i]) + '" selected="selected">' + escape_html(label) + '</option>');
         }
-        if (!options.length && !this.$input.attr('multiple')) {
+        if (!options.length) {
```
pedrocarmona: Either were, or in controller append null to all missing smart variables

app/views/blazer/_variables.html.erb
```diff
@@ -8,27 +8,29 @@
...
-            create: true
-          });
-        </script>
-      <% else %>
-        <%= text_field_tag var, params[var], style: "width: 120px; margin-right: 20px;", autofocus: i == 0 && !var.end_with?("_at") && !params[var], class: "form-control" %>
-        <% if var.end_with?("_at") %>
+      <div class="form-group">
+        <%= label_tag var, var %>
+        <% if (data = @smart_vars[var]) %>
+          <%= select_tag var, options_for_select([[nil, nil]] + data, selected: params[var]), style: "margin-right: 20px; width: 200px; display: none;", multiple: true %>
```
pedrocarmona: multiple: true could be conditional and based in regex from query or if variable is in plural (```user_ids```)

pedrocarmona:
@ankane, can you review?

bernstein7:
@pedrocarmona hey, why it was closed?

pedrocarmona:
it was a bit complex, I stopped enforcing a use case to it, but if you want I can restore the branch and we can test and improve together

Issue #145, opened 2017-04-16
Checks: support slack notifications instead of email


skatkov:
Currently, it's a popular practice to replace emails with slack-notifications in certain channels. Would be nice if Blazer could support slack notification as an option.

Our company currently has a very hacky version of blazer that support slack notifications. I'll try to find some personal time to convert it to be more general use.

ankane:
Added in #171

PR #146, opened 2017-04-18
Improve slack implementation

FanaHOVA:
As Stas was saying here (https://github.com/ankane/blazer/issues/145) I was working on adding Slack support for Blazer checks. Right now it works by reading Webhook URL, channel and username from ENV variables but obviously that's not optimal. I've placed the notifier in the `mailer` folder so that it'd be automatically loaded by Rails apps. 

It'd be nice to hear suggestions about how to implement this further; I was thinking of adding Slack options for each check in the same way that emails do, but each channel requires a specific webhook URL, so maybe a default channel would be best so the webhook can be hidden in `.env`. I was thinking about the ability to append to the message as well, for example to add Slack `@s` of people who should be interested in it.

### DIFF
--- a/app/mailers/blazer/check_mailer.rb
+++ b/app/mailers/blazer/check_mailer.rb
@@ -32,7 +32,7 @@ module Blazer
       notifier = Slack::Notifier.new ENV.fetch('WEBHOOK_URL'), channel: '#info-production',
                                                                username: 'Blazer Notifier'
 
-      notifier.ping "Check #{name} failing. @etavenn @skatkov @channel"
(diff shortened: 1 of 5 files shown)


FanaHOVA:
Got a cleaner implementation of this. Gonna reopen a PR after I clean out company @s from notifs!

blairanderson:
@FanaHOVA have you PR'd this?

FanaHOVA:
@blairanderson whoops! I forgot about this. I should still have the branch
somewhere locally, Hacktober is a good time to pick it back up I guess!

PR #147, opened 2017-04-19
Added pie chart

resool:
Pie chart is created when query returns two columns - string and numeric.


davidhrbac:
@resool Doesn't work for me...

resool:
@davidhrbac, checkout my fork here https://github.com/resool/blazer (my changes haven't been merged with blazer).

You can also enforce the type of graph you want to use by adding additional column named `_blazer_chart_type` and as a value one of these keywords: `bar`, `bar2`, `line`, `line2`, `table`, `pie`, fe:
```
   SELECT *,
          'bar' as _blazer_chart_type
     FROM results
```

Columns starting with `_blazer_` are not visible in blazer's results

davidhrbac:
@resool thaks, my fault. What's the reason for not merging this PR?

resool:
imo the pie chart shouldn't be the default because the only situation when you might need it is when the whole pie represents something, fa 356 days, 100% of subs etc.
Also I don't thing that creating special column in sql query to enforce graph type is a good practice. maybe it would be better to add graph_type column to query model?

resool:
+ it's the breaking change :)

cycomachead:
> maybe it would be better to add graph_type column to query model?

I've been considering this as well. It would be useful for some of my charts.

davidhrbac:
@ankane we would welcome pie chart and hidden columns...

Issue #56, opened 2016-04-20
Creating Public / Sharable Pages


cycomachead:
This one comes from Heroku's "Dataclips" feature. 

I'd like to _optionally_ be able to designate a dashboard / query as 'public' which would allow the page to be visited even if there's a `before_filter` active. This would make sharing some analysis easy, without needing to give users full admin access.

Bonus points that'd be cool:
- The CSV export URL could be public. (This would allow embedding CSV data in a Google Sheet, which is also a neat little trick. Heroku Dataclips does this.)
- Make the public pages read only if not logged in.
- Make the public pages not indexable by google

blairanderson:
design:
- `allow_public` config, default to false
- add column `public` boolean to queries, default to false
- add column `obscure_id` to queries, on create with SecureRandom
- add route to `ExportsController` if allow_public config is true
- respond to formats CSV/JSON

FanaHOVA:
@cycomachead @blairanderson not sure if you guys are still interested in this, giving it a shot in this PR, hoping to have it done by the end of the week! https://github.com/ankane/blazer/pull/153

ankane:
Cleaning up issues. For future readers, see https://github.com/ankane/blazer/pull/153#issuecomment-451785361.

PR #153, opened 2017-06-13
Add Public/Shareable queries

FanaHOVA:
Refers to issue https://github.com/ankane/blazer/issues/56

WIP PR. Missing:
- Allow multiple endpoints for export (JSON, CSV, etc)
- Don't allow Query edit while accessing a public page when not logged in OR when the user hasn't proved HTTP authentication details.

### DIFF
--- a/app/controllers/blazer/exports_controller.rb
+++ b/app/controllers/blazer/exports_controller.rb
@@ -1,0 +1,16 @@
+module Blazer
+  class ExportsController < BaseController
+    before_action :set_query
+
+    def show
+      return head :forbidden unless @query.public
+      redirect_to queries_path(query_id: @query.id, format: "csv"), params: {statement: @query.statement}
+    end
+
+    private
+
+    def set_query
+      @query = Blazer::Query.find_by(pid: params[:pid])
+    end
+  end
+end
--- a/app/controllers/blazer/exports_controller.rb
+++ b/app/controllers/blazer/exports_controller.rb
@@ -2,9 +2,12 @@ module Blazer
   class ExportsController < BaseController
     before_action :set_query
 
+    include ::Blazer::QueryRunner
+
     def show
       return head :forbidden unless @query.public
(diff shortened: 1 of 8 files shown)


ewalk153:
@FanaHOVA - I need this feature as well, would be happy to collaborate on it! I just pull down your branch. ideally, the `Blazer::ExportsController` would not redirect to the queries page, but instead just render the content from this controller.

In my case, I don't want to expose the private database `Query.id` attribute. Would you mind if I add a commit to your branch to experiment with this?

FanaHOVA:
@ewalk153 That's totally fine! It's been a really packed month and work and got little time for OS :( I was thinking of using `SecureRandom.uuid` to expose a public ID for each query since it'd mean no additional dependencies. What do you think?

ewalk153:
100% agree.

ewalk153:
It's not pretty, but I think this works.

https://github.com/Flatbook/blazer/tree/ew/public-sharing.

Wanna give it a shot?

FanaHOVA:
@ewalk153 like @xymbol was saying I think the `included` block isn't necessary. That's a hook called when you include the concern in the controller, so defining them in there doesn't really change much other than probably slowing it down a bit (I have no proof to back this up, just a gut feeling, callbacks/hooks always scare me :P). Other than that, looks good!

ewalk153:
I just tested it. I narrowed the `included` block down to just wrapping the `helper_method :blazer_time_value` line.

ewalk153:
Thanks @FanaHOVA and @xymbol for the feedback.

FanaHOVA:
@ewalk153 great; shouldn't you point your PR to my branch so we can merge it together and then ask @ankane to review? 🙂

ewalk153:
I'll make the PR!

ewalk153:
PR made: https://github.com/FanaHOVA/blazer/pull/1

FanaHOVA:
@ankane can you take a look at this? I merged @ewalk153's PR into mine, would be good to get some pointers from you with what changes are needed (if any)

PR #155, opened 2017-06-27
Compatibility with squeel

markrebec:
This is a small change to support compatibility with the [squeel](https://github.com/activerecord-hackery/squeel) gem while still maintaining compatibility with raw activerecord.


markrebec:
We finally ditched squeel, which also doesn't support rails 5, so this PR probably isn't worth leaving hanging. The commit should be easy to pluck though if you ever want it!

PR #156, opened 2017-07-04
Default presto schema

kwent:
The default presto schema is named `default` and not `public`


ankane:
Hey @kwent, thanks for the PR 👍  From my experience, Presto doesn't seem to have a default schema. Can you explain where/how to find this?

kwent:
There is a schema called "default" when I tried with AWS EMR at least.
Other examples and tutorial I can find use "default" as well as
http://diethardsteiner.github.io/big/data/2016/02/14/Presto.html.

Regards

ankane:
From what I can tell, it varies by connector, so don't see a strong reason to change.

PR #162, opened 2017-08-25
Specify where to define before_action method in README under Authenti…

justinfaulkner:
…cation > Other

I couldn't figure out from the docs how to use `before_action` in `blazer.yml`. With help from @cmckni3 on [stackoverflow](https://stackoverflow.com/questions/45849750/custom-authentication-for-blazer-rails), defining the custom auth method in `application_controller.rb` seems to work, but I'm not entirely sure that's what was intended.

If it is, I've taken a shot at adding to the docs on where to put the auth method.

Thanks for your work on blazer. It's amazing.

Justin

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,14 @@ Specify a `before_action` method to run in `blazer.yml`.
 ```yml
 before_action: require_admin
 ```
+Then define the custom authentication method in your `application_controller.rb`.
+
+```ruby
+def require_admin
+  # depending on your auth, maybe something like...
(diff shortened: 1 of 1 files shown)


cmckni3:
Looks good to me

ankane:
Looks great, thanks @justinfaulkner 👍

ankane:
In case someone comes across this in the future, the `before_action` should have a redirect or render, not return true or false. I missed this on the initial review.

cc #186

PR #163, opened 2017-09-04
Specify rails version

stefan-lz:
I'm working on a bit of a crusty rails (internal) app stuck on version 3, and I was hoping to use blazer for some simple reporting but realized it wouldn't work when I got this error.
```
Routing Error

undefined method `skip_action_callback' for Blazer::BaseController:Class
Did you mean?  skip_callback
```

So this pull request is to let everyone know it only works on rails 4 and above and don't go down the same path I did. :grimacing:

### DIFF
--- a/blazer.gemspec
+++ b/blazer.gemspec
@@ -17,7 +17,7 @@ Gem::Specification.new do |spec|
   spec.test_files    = spec.files.grep(%r{^(test|spec|features)/})
   spec.require_paths = ["lib"]
 
-  spec.add_dependency "rails"
+  spec.add_dependency "rails", ">= 4"
   spec.add_dependency "chartkick"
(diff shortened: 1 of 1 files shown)


ankane:
Hey @stefan-lz, thanks for the PR 👍  Let's just update the gemspec. It should be pretty clear when someone tries to install.

stefan-lz:
Yeah no worries! 

The readme has been updated.

With the gemspec version, looking at this commit, https://github.com/rails/rails/commit/9d62e04838f01f5589fa50b0baa480d60c815e2c the `filter` to `action` rename happened in `4.0.0.beta1` so `>=4` should cover it I would think.

ankane:
Awesome, thanks @stefan-lz!

PR #171, opened 2017-10-05
feat(Checks): add Slack notifications

FanaHOVA:
Adding Slack notifications for checks. 

Setup:
- Get a Webhook URL from Slack and set it as an ENV variable
- When creating a check, specify your Slack channel
- Receive Slack notifications at the same pace as email notifications

### DIFF
--- a/app/controllers/blazer/checks_controller.rb
+++ b/app/controllers/blazer/checks_controller.rb
@@ -46,7 +46,7 @@ module Blazer
     private
 
       def check_params
-        params.require(:check).permit(:query_id, :emails, :invert, :check_type, :schedule)
+        params.require(:check).permit(:query_id, :emails, :invert, :check_type, :schedule, :notify_slack)
       end
 
       def set_check
--- a/app/mailers/blazer/slack_notifier.rb
+++ b/app/mailers/blazer/slack_notifier.rb
@@ -6,21 +6,20 @@ module Blazer
 
     def self.state_change(check, state, state_was, rows_count)
       message = "#{check.query.name} check changed status from #{state_was} to #{state}. It now returns #{rows} rows."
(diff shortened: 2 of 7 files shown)


luddement:
Hello!

Not sure if the 'slack-notifier' gem is necessary. Maybe it is better to remove it and just do a plain request. Something like:

```
  def notify_slack(message, channel)
    slack_url = ENV['WEBHOOK_URL']
    return unless slack_url.present?

    params = {
      text: message,
      channel: channel
    }

    uri = URI.parse(slack_url)
    http = Net::HTTP.new(uri.host, uri.port)
    http.use_ssl = true

    request = Net::HTTP::Post.new(uri.request_uri)
    request.body = params.to_json
    http.request(request)
  end
```

FanaHOVA:
@luddement yea I was thinking about that. I wanted to leave the notifier in because people might want to hack on it and build cooler stuff on top of it, but it's probably better to keep dependencies light since it might not be needed for most people. Give me a couple mins to fix it up!

FanaHOVA:
@luddement realized that straight webhooks don't support changing channel for each notification, so I've swapped it to a checkbox. I guess most people will want notifications in the same channel, so it shouldn't be a big deal. If there's a request for it, we can allow that in the future.

CCing @blairanderson since he asked about this and obviously CCing @ankane for help in moving forward with this :)

feliperaul:
This would be really nice !

ankane:
Hey @FanaHOVA, big thanks for the PR 💯 and sorry it took so long to get to. This is great!

I think it's best to support multiple channels, so went that route. Added in https://github.com/ankane/blazer/commit/756d9156dd2e9bf870739a161a39d4e26fadcf75.

Issue #167, opened 2017-09-20
Linked columns for Rails nested routes


ibrahima:
The linked columns feature is great, but for Rails apps using nested routes, you often can't generate a URL with just the database ID - you often need other associations' IDs in the paths (eg. `/books/42/chapters/16`). Is it possible to incorporate those other path segments into links? I.e., I'd like to be able to make links like `"/books/{book_id}/chapters/{chapter_id}"` instead of just "/chapters/{value}".

I think in general if links could include arbitrary columns in the format string, that might be sufficient - in most of these cases I think you'd have to do a join to get the parent resource IDs anyway. This could also be done through something like smart columns so that you don't have to explicitly join those parent resources, but it'd have to be different since you wouldn't want to see those extra values in general.

This might be a bit convoluted, but it'd be really helpful in many Rails apps!

ibrahima:
Aaand just like last time I posted a feature request (#151) I realized that with judicious SQL querying, you can achieve this. In this case, if you use PostgreSQL (and possibly other SQLs), you can use the `format` function to generate a URL, peforming the joins yourself:

```SQL
SELECT format('https://mydomain.com/books/%s/chapters/%s', chapters.book_id, chapter_id)
FROM bookmarks
INNER JOIN chapters ON chapters.id = bookmarks.chapter_id;
```

Where say, the bookmarks table includes chapter_ids, and then you join on chapters to get the book_id.

Still would be nice to use linked columns so that the whole URL doesn't have to be in the column though, but that's fairly minor! And it'd be nice to have that capability of defining the joins once in a similar way to smart columns so that you wouldn't have to manually write joins and format strings each place such an ID comes up. But this solves my immediate problem!

PR #173, opened 2017-10-14
Linked columns with row values

ibrahima:
Implementation for #167 (and more). Basically, this change allows using any row values in linked column templates, so that if you do have nested rails resources, as long as you can somehow make sure that you have the right columns in your result table, it'll generate the links for you.

Thoughts:

- if you don't have all the necessary columns, should it not generate a URL, or just generate a bogus URL? To me it seems like it'd be best to not attempt to generate a URL if there are any interpolated values that don't exist in the row, but I haven't implemented that yet (this was mostly a small proof of concept)

- Is it a performance or security issue to gsub all column names, rather than looking for the specific values that are included in the template string? It might be slightly more efficient to first scan the string for segments that look like "{\w+}" (for instance) and then look those up in the columns list if any are found, but it seems like either way you're iterating through the columns list, and this way we need a regex which is potentially error prone. The current implementation is straightforward and just works, even if it's searching the format string N times for each column, times the number of rows. Although if you have more columns with link templates it could start to build up... but I kinda think thinking about this is premature optimization.

- I've left in the current behavior of interpolating `{value}` for the value for that column, seems useful enough to keep both options.

Any feedback is greatly appreciated!

### DIFF
--- a/app/helpers/blazer/base_helper.rb
+++ b/app/helpers/blazer/base_helper.rb
@@ -28,6 +28,10 @@ module Blazer
       end
     end
 
+    def build_column_link(template, value)
+      template.gsub("{value}", url_encode(value.to_s))
+    end
+
     def blazer_maps?
       ENV["MAPBOX_ACCESS_TOKEN"].present?
     end
--- a/app/helpers/blazer/base_helper.rb
+++ b/app/helpers/blazer/base_helper.rb
@@ -28,8 +28,12 @@ module Blazer
       end
     end
 
-    def build_column_link(template, value)
-      template.gsub("{value}", url_encode(value.to_s))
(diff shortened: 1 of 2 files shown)


ankane:
Hey @ibrahima, thanks for the PR 👍 I've thought about this before as well, but ultimately decided against it, since it adds complexity (like the issues you mentioned) and you can generate full urls through concatenation that are linkable when needed.

ibrahima:
Well, I still feel like it'd be useful because concatenation is pretty annoying to have to repeat all the time, but we'd definitely have to resolve some of those issues before something like that could be production ready. I just figured that it'd be nice to just configure those URLs in one place, and then people writing new queries don't have to figure out how to form a URL.

PR #182, opened 2017-12-23
No more to lose unsaved SQL when refreshing "queries/new" page

ilstar:
## What

* Cache SQLs on "queries/new" page while you're typing, so you won't lose your queries if the browser get closed (or go back to previous page) accidentally.
* Multi-tab is also supported, so the data won't be mixed up even there are multiple tabs.

Note:
* Since it uses `sessionStorage`, we can't restore the data if you open a new tab, you have to restore the data in the original tab.

### DIFF
--- a/app/assets/javascripts/blazer/application.js
+++ b/app/assets/javascripts/blazer/application.js
@@ -16,6 +16,7 @@
 //= require ./routes
 //= require ./queries
 //= require ./fuzzysearch
+//= require ./draft_store
 
 Vue.config.devtools = false
 
@@ -76,4 +77,3 @@ function preventBackspaceNav() {
 }
 
 preventBackspaceNav()
-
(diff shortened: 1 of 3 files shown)


ankane:
Thanks @ilstar 👍 Can we try this out on the Instacart fork for a while and see how it feels before adding to the open-source version? One thing we may want to do is expire drafts after 5 or 10 minutes so you don't have to clear the window when you try to create a new query an hour later. Also, you probably want to clear the draft when the query is saved.

ankane:
Awesome, thanks @ilstar 💯

ankane:
Hey @ilstar, thinking about it more, I think there's a slight security implication with this approach. Someone on the same computer can read draft queries even after a user logs out. I don't thing there's a great way to handle this without closer integration with authentication. This may seem small but think I need to revert because of this. Sorry I missed this the first time around. I may reconsider in the future.

ilstar:
That’s a good catch, thanks for taking care of it!

PR #187, opened 2018-02-15
depend on railties instead of rails

jjb:
to facilitate the user including only the rails librarites
their project needs (e.g. including everything but actioncable)

let me know what you think!


ankane:
Looks good, thanks @jjb 👍

jjb:
🆒

PR #188, opened 2018-02-27
Make Blazer replica lag aware

rnaud:


ankane:
Thanks @rnaud 👍 What's the motivation for adding a new replica lag state? (get check failing emails when lag is bad?)

rnaud:
@ankane sorry for the delayed answer. Right now when we get hit by large replica lag, a lot of our checks hard fail.

For example, the check "Was there an order placed in the last 5 minutes on every platform?" will fail if we have more than 5 minutes of lag. 

We want to make sure that blazer is lag aware so that we can report on lag in the same way we report on checks timing out: orange state.

ankane:
Makes sense. The current implementation adds the lag query to every run. I think it'd be better to only incorporate this functionality into checks. For instance, before a check runs, check the lag before running the query. Also, don't think there should be a default - think this should only take effect if an option is set in the config for that database (maybe `max_check_lag`).

Also, since check states can be updated when a query runs in the web interface, we'll need to handle that case as well (with the new approach).

ankane:
Cleaning up stale PRs

PR #197, opened 2018-06-08
Feature: Automatic conversion of enum values

haffla:
Hello,

this PR makes it possible to automatically convert column integer values to the corresponding string values for enum types. The column name needs to be in a special form to enable this: name_of_class_enum_name_of_field.

So if you have a class User with an enum field `state` you could write a query such as

`SELECT state as user_enum_state, COUNT(*) FROM users GROUP BY 1;`

Here all the integer values would be converted to the corresponding string values resulting in a more readable column chart.

I wanted to find out if this feature is desirable before I write the docs.

### DIFF
--- a/lib/blazer/result.rb
+++ b/lib/blazer/result.rb
@@ -9,6 +9,7 @@ module Blazer
       @error = error
       @cached_at = cached_at
       @just_cached = just_cached
+      convert_enum_values
     end
 
     def timed_out?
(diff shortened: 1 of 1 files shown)


ankane:
Hey @haffla, thanks for the PR 👍 I feel like this is a bit too much magic. I'd recommend smart columns for translating integer enums to strings.

haffla:
Hey @ankane, yeah I totally understand. I don't have a good feeling with this PR either.

Issue #135, opened 2017-03-24
Change default time range for start_time & end_time


marcusg:
Is it possible to change the default values for the datepicker values when entering the dashboard? For example: last year instead of last 30 days. Thanks!

ankane:
Hey @marcusg, there's no setting for this right now.

marcusg:
@ankane ok, thanks for the response. Do you think it would be an useful option?

ankane:
I don't personally have a strong need, but can see it being useful to others, so happy to accept a PR if it's done in a clean way.

ankane:
Going to close this since discussion is stale, but previous comment still applies.

PR #200, opened 2018-07-17
Add yesterday button to daterangepicker

Pazzik:
In our team there is a request to add "yesterday" button. Maybe this will be useful to someone else.


pedrocarmona:
maybe you could make it as a ahoy config param (https://github.com/ankane/blazer/blob/master/app/views/blazer/_variables.html.erb#L47)

ankane:
Hey @Pazzik, thanks for the PR 👍 I'm open to making this configurable as mentioned in #135, but don't want to add any more static ranges.

PR #208, opened 2018-09-11
Fix for BigQuery adapter with version 1.8

popcorn4dinner:
When using blazer with the newest version of `google-cloud-bigquery`, running queries results in `undefined method 'complete?' for #<Google::Cloud::BigQuery::Data ...`.

This change attempts to fix the BigQuery adapter.


ankane:
Thanks @popcorn4dinner 👍

Issue #203, opened 2018-08-03
App wide CSP breaking Blazer


Masa331:
Hi,

first of all, thanks for this great gem!

I have a minor problem with CSP. I have it set up very strictly site-wide in my app. I only allow 'self' and 'https' directives which break Blazer functionality because it needs 'unsafe_inline' and others. I'm fine to relax it for Blazer but i'm not sure there is a way to do it from my app(without hacks). However, it could be done directly in Blazer. I'v quickly tried it in my fork and it works allright.

I can prepare pull request if you like it or tell me please if there is another preffered solution?

Here is the code: https://github.com/ankane/blazer/compare/master...Masa331:master

It could be moved to BaseController or something but this was just a quick try..

Have a nice day, Masa331

ankane:
Hey @Masa331, thanks for sharing 👍 I'm hesitant to have Blazer automatically override app-wide CSP by default. Maybe we could add an option like `override_csp` for it? Also, we'd need to set the CSP header directly so it works with previous versions of Rails.

Masa331:
Hi @ankane,

i understand. It also might be enough to mention CSPs in readme and describe and show what's necessary for Blazer to properly work. I think this was in fact the biggest problem when i was dealing with it - at start i didn't know why it's not working and i had to search for it elsewhere.

What do you think? I can prepare the feature with `override_csp` option or write the docs if you want. Both seems good solutions to me

Masa331

ankane:
I think adding an option for it will take nearly the same amount of time as documenting it (and won't require user maintenance if it changes), so that would be my preference.

Masa331:
ok, will do

ankane:
Fixed in #209

KjellMorgenstern:
The override does only set the default-src , but not script_src or style_src. Also, it doesn't seem to work with SecureHeaders. I therefore applied the following override (https://guides.rubyonrails.org/engines.html#overriding-models-and-controllers) 

```
Blazer::BaseController.class_eval do
  after_action do
    append_content_security_policy_directives(
      style_src: %w( https://api.mapbox.com/mapbox.js/ ),
      script_src: %w( https: 'unsafe-inline' 'unsafe-eval')
    )
  end
end
```

Maybe this could be simplified if blazer.yml was actually a blazer.rb initializer?

Instead of
`override_csp: true `
it would then accept
```
override_csp: {
  ## insert block with csp override command here
  append_content_security_policy_directives(
[...]

PR #209, opened 2018-09-16
Add option to override app-wide csp settings

Masa331:
Hi @ankane,

this code adds Blazer option to override strict app csp which might block Blazer's functionality as discussed in #203 .

The csp headers are set directly through `response.headers` in `after_action`. So it should work with Rails < 5.2 also. However for Rails < 4 there would have to be branch which would use `after_filter` instead of `after_action`. But i don't know how much old Rails you want to support?

Have a nice day,
Masa331


ankane:
Looks great, thanks @Masa331 👍

PR #210, opened 2018-09-17
Add flash message to notify of disallowed FERPA queries

ctoynbee:


ctoynbee:
Apologies. I was suppose to close that. Was meant for internal use only. Thanks for a great library!

PR #218, opened 2018-10-25
Newell/timelines

andrew-newell:
I love Blazer but it was missing something for me and I felt vis.js timelines provide the kind of data visualization I need to make Blazer an essential piece of kit.

Here's an example of a vis.js timeline http://visjs.org/examples/timeline/basicUsage.html

Here's a screenshot of a grouped timeline produced by Blazer https://i.imgur.com/pAvTPNj.png


andrew-newell:
Upon further inspection it looks like there are a few more html/js/css things to work out so this plays nicely with dashboards.

ankane:
Hey @andrew-newell, thanks for the PR 👍 However, I don't plan to support any more chart types right now.

Also, as a general open source best practice, avoid bumping version numbers in PRs. Maintainers almost always have their own process for this.

andrew-newell:
Ah, that's a shame, well thanks for all your effort so far. I'll keep my eyes on this repo in case you ever change your mind. Also, thanks for the tip. ☮️

PR #221, opened 2018-12-13
Updated to add information on admin checks

rathboma:
Added an example using devise and rails to authenticate users before giving them access.


ankane:
Hey @rathboma, thanks for the PR 👍 This example is already in the readme.

PR #222, opened 2018-12-14
Use assets path from yml file

charusat09:


ankane:
Hey @charusat09, what the use case for this?

charusat09:
@ankane We have custom routes for assets. I was not able to use this gem since default `/assets` path is not accessible through our routes. So, I make it configurable so that, anyone with different settings for routes can use this gem. Thanks.

ankane:
Unfortunately, I don't think this is common enough to include an option for.

charusat09:
Ok. Make sense. Anyways, I have stared using my forked version.

PR #223, opened 2018-12-14
Add MariaDB support

ArtOfCode-:
MariaDB is almost identical to MySQL, but uses a different system variable for query timeouts.


ankane:
Thanks @ArtOfCode- 👍 Updated to use the built-in Rails method for MariaDB. https://github.com/ankane/blazer/commit/9d76b3b56f35e36116f4f5838f30230fc4a2c2e9

ArtOfCode-:
Ah, missed a trick. Nice :)

PR #224, opened 2018-12-20
add bash and docker

pedrocarmona:
Hi @ankane,

I had this ideia to make a `bash_adapter` and `docker_adapter`.
The main goal is to allow to use other languages via the CLI. 

simple python script, as long as it prints json:
```
import json

array = [
 {"someKey_1":1, "someKey_2":2}, 
 {"someKey_1":3, "someKey_2":4}
]
print(json.dumps(array))
```

You can also use env variables:
```
import json
import os

array = [
 {"ENV": os.getenv('FOO')}
]
print(json.dumps(array))
```

Moreover, one can fetch data from multiple sources and combine it in a python pandas script, then output the dateset as json that gets parsed and displayed by blazer.


ankane:
Hey @pedrocarmona, this is a neat feature and the code looks great 👍 I'm worried about the security implications for the bash adapter (can run arbitrary code on the server) and don't think either will be commonly used. I imagine most people interested in this functionality are better off with a Jupyter notebook.

PR #231, opened 2019-02-18
Colors of pie chart for more then 20 records are missing - Issue Fixed

nabeelnazir:
**Problem**
When we write some query and if the database returns more then 20 records then for the 21st and all upcoming records colors won't show up on the pie chart. 

**Solution**
I have added a small patch which will generate an array of random colors and passed it to the pie chart as chart options. The size/length of colors array will be equal to the number of records returned from the database in the query.
This solution will assign colors to every section of the pie chart and it won't miss any color. 

**Example**
Attached are the screenshots of the problems and a screenshot after the solution applied. 

Problem 1:
problem number 1
Problem 2:
problem number 2
Solution:
after solution applied


muhammadumarkhan:
Have the same issue, @ankane  please merge this.

ankane:
Cleaning up PRs

PR #232, opened 2019-02-20
Idea: use liquid

pedrocarmona:
This PR introduces liquid as main template engine for queries. Allows to write queries with null values, and also we could easily refactor the query below in the images to be IN ({{ occupation_id }} | join(', ') ), and then allow selectize to select multiple.

To do if idea accepted
- [ ] create migration to update `{occupation_id}` to  `{{ occupation_id }}` 

screenshot 2019-02-20 at 21 22 17
screenshot 2019-02-20 at 21 22 53
screenshot 2019-02-20 at 21 22 59

### DIFF
--- a/app/controllers/blazer/base_controller.rb
+++ b/app/controllers/blazer/base_controller.rb
@@ -34,45 +34,46 @@ module Blazer
 
     private
 
-      def process_vars(statement, data_source)
-        (@bind_vars ||= []).concat(Blazer.extract_vars(statement)).uniq!
-        @bind_vars.each do |var|
-          params[var] ||= Blazer.data_sources[data_source].variable_defaults[var]
+      def fix_var_value(var, value)
+        if ["start_time", "end_time"].include?(var)
+          value = value.to_s.gsub(" ", "+") # fix for Quip bug
         end
-        @success = @bind_vars.all? { |v| params[v] }
-
-        if @success
-          @bind_vars.each do |var|
-            value = params[var].presence
-            if value
-              if ["start_time", "end_time"].include?(var)
-                value = value.to_s.gsub(" ", "+") # fix for Quip bug
-              end
-
-              if var.end_with?("_at")
-                begin
-                  value = Blazer.time_zone.parse(value)
-                rescue
-                  # do nothing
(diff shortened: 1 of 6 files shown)


pedrocarmona:
in the example, another case could also be that if nothing is selected, then the where statement is not used

ankane:
Hey @pedrocarmona, this is an interesting idea for writing more advanced queries, but I feel like this could be accomplished in a different way. For multiple selects, I think we can follow the current convention and use something `IN {var}` to generate one. For null, we could do something similar with `= {var}` and substitute `IS NULL`.

pedrocarmona:
I want to use this as part of a generic dashboard. I havent made it yet. 
The goal is to allow to magically group by multiple columns in order to drill down the data, but also allow where statements.
I see the group by also as smart_column of multiple values {week, month, columnA, columnB, etc}.

When giving flexibility to select by a specific period, the query becomes a bit more complex. Given a select with a group by period { day, week, month, etc}, then in the column selected is `date_trunc('month', created_at) as month`, but in column in the group by is ` date_trunc('month', created_at)` without the `as month`. This is something specific to the query, and that could be customized with liquid conditions.

```
select 
  date_trunc('month', created_at) as month,
  columnA,
  count(*)
where 
  columnB IN ('1','2','3')
group by date_trunc('month', created_at), columnA
```

In my opinion, empowering the queries with liquid gives enough flexibility to reuse one query for multiple cases, allowing to drill down data, and create new dimensions (conditionally) that dont exist in the underlying data storage.

ankane:
I'm still not sure of the practical benefit of the added complexity, but if you want to build out your vision in a fork and try it out for a while, I can take a look at a later time.

pedrocarmona:
Sure. Thx Andrew. I will close for now.

pedrocarmona:
Hi @ankane, 

In order to have generic dashboards, I build this "DSL" to create dashboards, with predefined dimensions (things that you can group) and measures (fields that we can sum, count, avg, etc). 

I still have few things to adjust, this is the sneak preview: https://github.com/pedrocarmona/tailor_made

Issue #234, opened 2019-03-06
Athena Adapter - client_request_token hash does not allow you to run same query multiple times


preetsethi:
Before I say anything else: wanted to say thanks to @ankane and all the other contributors for all the hard work on this library, it's kick ass!

I've uncovered some interesting behavior between Athena and the Blazer AthenaAdapter.
https://github.com/ankane/blazer/blob/master/lib/blazer/adapters/athena_adapter.rb#L16
```ruby
client.start_query_execution(
  query_string: statement,
  # use token so we fetch cached results after query is run
  client_request_token: Digest::MD5.hexdigest(statement),
  query_execution_context: {
    database: database,
  },
  result_configuration: {
    output_location: settings["output_location"]
  }
)
```

Using `client_request_token: Digest::MD5.hexdigest(statement)` is a smooth solution to track / fetch query results. 

However, only using `statement` as an input to the hash is causing issues. 

I believe this is because `client_request_token` is unique at the account level in Athena. There's not a lot of details on the scope but here's the relevant section of the docs:
https://docs.aws.amazon.com/athena/latest/APIReference/API_StartQueryExecution.html#API_StartQueryExecution_RequestSyntax
```
ClientRequestToken
  A unique case-sensitive string used to ensure the request to create the query is idempotent (executes only once). 
  If another StartQueryExecution request is received, the same response is returned and another query is not created.
  If a parameter has changed, for example, the QueryString, an error is returned.
```

This problem is two fold:
1. I have the same tables with the same names across two Athena / Glue databases. I set up two data sources in Blazer, one for each database. I'm able to run the statement on one data source / database, but when I try to run the statement on the second data source / database, I get the following error:`Idempotent parameters do not match`. **This is particularly annoying when trying to preview tables.** Going back to the AWS documentation, you'll see this line: `If a parameter has changed, for example, the QueryString, an error is returned.` Since the `client_request_token` across the two data sources is the same but the `query_execution_context: {database: database}` has changed, it throws the `Idempotent parameters` error.

Solution: 
Add `database` as an input to the hash. The `client_request_token` hash must factor in the `database`. 
-or-
Ensure the preview table statement uses the fully qualified database name in the query itself. Ex: `select * from 'database'.table` vs `select * from table`. 
In either case, this one seems pretty straightforward, pretty clean.

[...]

PR #235, opened 2019-03-08
Add data source id to digest for AthenaAdapter

preetsethi:
Quick fix for https://github.com/ankane/blazer/issues/234

Tested this in production via a custom adapter in my application. Using this to resolve my issue until PR is merged.

@ankane 
If you're not fond of the inline array join, or you foresee adding other inputs to the digest, let me know. Open to whatever.


ankane:
Thanks @preetsethi. Think we want to use `data_source.id` instead of `data_source`. Also, re join: Blazer uses `join("/")` for other cache keys.

preetsethi:
@ankane Updated

PR #236, opened 2019-03-19
Bug Fix: StupidTable Plugin - fix issues with sorting integers and floats in the results table

basilkhan05:
When clicking on the results table headers to sort, the integers and floats are not sorted correctly or as expected. 
- This is because the commas in the integers and floats, added by the plugin, break this functionality with `parseInt` and `parseFloat`.
- `removeCommas` method was added to remove the commas prior to using `parseInt` and `parseFloat`.

The video below reproduces the issue and shows the fix in action:

[Blazer PR | Bug Fix for results table sorting](https://watch.basilkhan.ca/watch/ywLm1cGS5XnMS5mQnLQzmk)

### DIFF
--- a/app/assets/javascripts/blazer/application.js
+++ b/app/assets/javascripts/blazer/application.js
@@ -1,6 +1,7 @@
 //= require ./jquery
 //= require ./jquery-ujs
 //= require ./stupidtable
+//= require ./stupidtable-custom-settings
 //= require ./jquery.stickytableheaders
 //= require ./selectize
 //= require ./highlight.min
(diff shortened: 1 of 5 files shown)


ankane:
Hey @basilkhan05, thanks for the PR 👍 I think it'd be better to use a custom data type rather than modify the JavaScript file.

https://github.com/joequery/Stupid-Table-Plugin#creating-your-own-data-types

basilkhan05:
Thanks @ankane for the suggestion! Just pushed the revised fix with custom data types. Let me know if this works.

P.S. Thanks for this awesome gem ❤️

ankane:
Looks great, thanks @basilkhan05 👍

PR #237, opened 2019-03-22
Find `Blazer.user_method` if private or protected

trevororeilly:
Some authentication frameworks (e.g. Clearance) have a private
`current_user` method. Passing `true` as the second parameter to
`respond_to?` allows private and protected methods to also be
found.


ankane:
Awesome, thanks @trevororeilly 👍

Issue #24, opened 2015-10-06
Todo


ankane:
- [ ] keep params when editing dashboard
- [ ] hide mastermind if no user class
- [ ] handle single value on dashboards
- [ ] better navigation
- [ ] add to dashboard from query page
- [ ] better time defaults (4, weeks, 12 weeks, 52 weeks)
- [ ] a way to organize (categories)
- [ ] treat smart column as string for chart generation
- [ ] show error message on server error

Even Later

- [ ] default name for fork
- [ ] dashing integration (or replacement)?
- [ ] auto-refresh dashboards

chrisgogreen:
+ 1 for the download button

spdawson:
How about "smart" handling of columns/attributes that make use of `ActiveRecord::Enum`?

anshul:
Feature request:  A way to do pivot tables.

6thSigma:
This may be out of the scope of this gem, but the ability to add a moving average line to graphs would be awesome.

w0www:
up!

holden:
+1 for embeddable charts

fwolfst:
Idea: Next to schema (or even query), show erd (https://github.com/amatsuda/erd or https://github.com/voormedia/rails-erd) if respective gem is available. Btw are you okay with "lazy" idea contributions as comments in this thread?

pedrocarmona:
idea use liquid to build the sql queries:
- instead of  `{ smart_variable }`,  `{{ smart_variable }}` which would be inserted in the liquid context
- allows building dynamic sql with cycles.

cyruscavalcante:
smart variable -> chage null by '%' when nathing is selected on select box
ex:
https://blazer.dokkuapp.com/queries/1-smart-variable?occupation_id=
must result all occupations

nfedyashev:
idea:

When a user has lots of pins to display it would be really handy if you can just order records by date of creation and make the corresponding pin stand out when you hover over a certain line in the table below Mapbox area.

Actually, I've tried to implement it myself but from what I found it's not a trivial task and unfortunately I don't have enough time at the moment.

voltechs:
Allow for optional "schema" support. (`"blazer"."audits"`, `"blazer"."checks"`).

Would help "organize" domain-specific schemas as per "best practices" (not using the `public` schema):
https://www.postgresql.org/docs/current/ddl-schemas.html#DDL-SCHEMAS-PUBLIC
[...]

PR #238, opened 2019-03-25
Hide Mastermind column on home page if user class does not exist

basilkhan05:
This is a simple fix that removes the Mastermind column on the home page when either the default or user defined `user_class` does not exist.

Below is a quick video showcasing the changes in the PR:

[Play Video
](https://watch.basilkhan.ca/watch/NRZAh2wNcB7YkNm8ibBM7P)
[Blazer PR | Hide Mastermind if user_class does not exist](https://watch.basilkhan.ca/watch/NRZAh2wNcB7YkNm8ibBM7P)

This PR is in reference to the `- hide mastermind if no user class` from Issue #24

### DIFF
--- a/app/views/blazer/queries/home.html.erb
+++ b/app/views/blazer/queries/home.html.erb
@@ -32,7 +32,9 @@
     <thead>
       <tr>
         <th>Name</th>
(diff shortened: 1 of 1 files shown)


ankane:
Looks great, thanks @basilkhan05 👍

PR #244, opened 2019-05-28
Adds option to configure mapbox access token in settings

marcusmalmberg:
Adds option to configure mapbox access token as a setting. 

This is useful in projects using rails secrets instead of setting environment variables.

This shouldn't break anything since it defaults the setting to the environment variable.

### DIFF
--- a/app/helpers/blazer/base_helper.rb
+++ b/app/helpers/blazer/base_helper.rb
@@ -29,7 +29,7 @@ module Blazer
     end
 
     def blazer_maps?
-      ENV["MAPBOX_ACCESS_TOKEN"].present?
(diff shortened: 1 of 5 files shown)


ankane:
Hey @marcusmalmberg, thanks for the PR 👍  Looks good to me. I'd prefer not to change the readme to keep it consistent with other instructions (and as is, it somewhat looks like it suggests hardcoding the token). Would be good to add it to the installation template similar to `slack_webhook_url`.

marcusmalmberg:
@ankane Sure, is this better?

ankane:
Looks beautiful, thanks!

Issue #246, opened 2019-06-13
State change slack notification should not depends on emails presence


sandeep-patle1508:
Hi,
Why are we checking emails.present? for slack notification also?
https://github.com/ankane/blazer/blob/78609f3fbe8349a2333d720128f405ae1f8fb54e/app/models/blazer/check.rb#L70

As this email field is optional so we should check slack_channel.present? here for slack.

Thanks!

PR #247, opened 2019-06-13
Issue-246: State change slack notification should not depends on emails|

sandeep-patle1508:
Issue-246: State change slack notification should depends on channel presence instead of emails.
Issue - https://github.com/ankane/blazer/issues/246

### DIFF
--- a/app/models/blazer/check.rb
+++ b/app/models/blazer/check.rb
@@ -68,8 +68,13 @@ module Blazer
 
       # do not notify on creation, except when not passing
       if (state_was != "new" || state != "passing") && state != state_was && emails.present?
-        Blazer::CheckMailer.state_change(self, state, state_was, result.rows.size, message, result.columns, result.rows.first(10).as_json, result.column_types, check_type).deliver_now
-        Blazer::SlackNotifier.state_change(self, state, state_was, result.rows.size, message, check_type)
+        if emails.present?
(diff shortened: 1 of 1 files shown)


app/models/blazer/check.rb
```diff
@@ -68,8 +68,13 @@ def update_state(result)
 
       # do not notify on creation, except when not passing
       if (state_was != "new" || state != "passing") && state != state_was && emails.present?
```
ankane: Think you meant to remove the last condition.

app/models/blazer/check.rb
```diff
@@ -68,8 +68,13 @@ def update_state(result)
...
-        Blazer::CheckMailer.state_change(self, state, state_was, result.rows.size, message, result.columns, result.rows.first(10).as_json, result.column_types, check_type).deliver_now
-        Blazer::SlackNotifier.state_change(self, state, state_was, result.rows.size, message, check_type)
+        if emails.present?
+          Blazer::CheckMailer.state_change(self, state, state_was, result.rows.size, message, result.columns, result.rows.first(10).as_json, result.column_types, check_type).deliver_now
+        end
+
+        if slack_channels.present?
```
ankane: Better to use `split_slack_channels.any?` for backwards compatibility, as older installs won't have the `slack_channels` column.

PR #249, opened 2019-06-30
Add Support For Slack App

eran-barkai:
* Message get edited when test passing
image

* Add result table in ASCII
image

you need to define this permission for slack app: 

channels:read , chat:write:bot, chat:write:bot, groups:read, incoming-webhook,bot,
usergroups:read

image

BLAZER_SLACK_APP_TOKEN:
image

### DIFF
--- a/app/mailers/blazer/slack_notifier.rb
+++ b/app/mailers/blazer/slack_notifier.rb
@@ -2,6 +2,24 @@ require "net/http"
 
 module Blazer
   class SlackNotifier
+
+    class << self
+      attr_accessor :postMessageUrl
+      attr_accessor :updateUrl
+      attr_accessor :token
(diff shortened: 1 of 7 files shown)


ankane:
Hey @eran-barkai, thanks for the PR 👍 What's the reason for changing the Slack integration?

eran-barkai:
Webhook integration does not support updating existing messages. 
which it does now.

people in my organisation were complaining on the amount of messages. so this is one of the steps i took to reduce amount and clutter.

the other part of this jira is adding results to the slack msg. this allow us to understand the alert without diving in.

ankane:
I appreciate the PR, but think the current Slack integration is fine, and it requires a lot less permissions.

PR #255, opened 2019-08-08
Add expiration mechanism for AthenaAdapter's client_request_token

### LINKED ISSUE #254: Athena Adapter - client_request_token and caching
I raised a lot of these issues in #234. At the time, I didn't have a great solution.

However, as I have spent more time dealing with the problem:  the Athena adapter, in a few specific but common cases, is nearly unusable without some type of cache expiration for this token. 

Common use case: you have a dashboard query that you check periodically, which runs aggregates on time series data. Blazer makes it soo easy to turn this into a pretty chart too.
With Kinesis+Athena, you're doing date-time-based partitions and attaching incoming data as partitions.

The current hashing factors that produce the lookup token: it basically means I'm getting the cached results from when I wrote the query a week ago. I have to change the query in arbitrary ways like switch order direction... or even worse, add extraneous fields like dates or random strings, to generate a new `client_request_token`.

There must be a cache expiration mechanism for any given query, even if its a day, week, month, year, whatever.

I will have my proposal up for you as a PR in a few days. While I sense there will be objections to the manner in which I allow configurability, but I beg of you, we must add some sensible cache expiration factor to the `client_request_token` hashing mechanism.

To be continued ...

preetsethi:
- Adds a truncated Timestamp to the digest inputs for client_request_token
- Timestamp truncation is configurabile via adapter settings (See Time#change)
- Default expiration is set to a 24 hour window

resolves #254

### DIFF
--- a/README.md
+++ b/README.md
@@ -504,7 +504,12 @@ data_sources:
     adapter: athena
     database: database
     output_location: s3://some-bucket/
+    client_request_token:
+      expiration_config:
+        - !ruby/sym sec: 0
 ```
+See [Time#change](https://apidock.com/rails/Time/change) for details on configuring an expiration window for [Athena's `client_request_token`](https://docs.aws.amazon.com/athena/latest/APIReference/API_StartQueryExecution.html#API_StartQueryExecution_RequestSyntax
+).
 
 ### Amazon Redshift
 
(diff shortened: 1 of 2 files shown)


preetsethi:
@ankane Closing this, please see: https://github.com/ankane/blazer/issues/254#issuecomment-546680452

Once again, thanks man!

PR #257, opened 2019-08-08
Include initial row when column information is absent from Athena query response

### LINKED ISSUE #252: Athena Adapter - some queries incorrectly return 0 rows
### Problem
Specifically in the case of:
```
MSCK REPAIR TABLE example_table;
```
Was returning zero rows even in cases I was sure we just dropped fat logs.

Diving into the Athena Client response:
```
#<struct Aws::Athena::Types::GetQueryResultsOutput update_count=nil, result_set=
  #<struct Aws::Athena::Types::ResultSet rows=[
    #<struct Aws::Athena::Types::Row data=[
      #<struct Aws::Athena::Types::Datum var_char_value="Partitions not in metastore:\texample_partitioned_table:2019/08/07/14">
    ]>
  ], result_set_metadata=#<struct Aws::Athena::Types::ResultSetMetadata column_info=[]>>, next_token=nil>
```

In `AthenaAdapter`: 
When processing the result, problematic assumption here:
```
column_info = resp.result_set.result_set_metadata.column_info
```
Column info is not always present.

Thus, the initial index is off:
```
rows = untyped_rows[1..-1] || []
```
Zero index is in fact relevant data in this case.

### Solution
Plan to pivot on the presence of column_info and handle appropriately from there.
I'll submit a PR.

### Question
Do I add generic column names? Column 1, Column 2, Column 3, etc

preetsethi:
Resolves #252

### DIFF
--- a/lib/blazer/adapters/athena_adapter.rb
+++ b/lib/blazer/adapters/athena_adapter.rb
@@ -56,6 +56,7 @@ module Blazer
             utc = ActiveSupport::TimeZone['Etc/UTC']
 
             rows = untyped_rows[1..-1] || []
+            rows = untyped_rows[0..-1] unless column_info.present?
             column_types.each_with_index do |ct, i|
               # TODO more column_types
               case ct


ankane:
Thanks for this one as well 👍

PR #273, opened 2019-12-05
Support explicitly defining AWS credentials for Athena adapter

bobby-vandiver:
Allows the credentials for the Athena adapter to be configured explicitly. If the settings are not provided, the AWS clients fallback to the normal process for resolving credentials that is currently used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -506,6 +506,18 @@ data_sources:
     output_location: s3://some-bucket/
 ```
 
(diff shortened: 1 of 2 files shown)


ankane:
Thanks for this one as well! Overall, looks good. Will test it out when I have some time.

ankane:
Sorry it took so long to test. Thanks @bobby-vandiver!

PR #256, opened 2019-08-08
Rescue and retry on get_query_results QUEUED exception

preetsethi:
Resolves #253

PR #274, opened 2019-12-05
Poll Athena to determine when a query is being processed

bobby-vandiver:
The current implementation failed to retry queries that are in the `QUEUED` state prior to #256.

Instead of attempting to get the results of the query and retrying based on an error message which might change in the future, this PR checks the status of the query via the AWS APIs to determine when it has succeeded and the results can be retrieved.

This also includes a minor change to perform the creation of the request token in a separate method that can be easily monkey patched for applications that always want to execute a query. Currently if someone needs this behavior, they must duplicate the entirety of the `run_statement` method to support this.

**EDIT**: Note this was originally based on a version of Blazer prior to #256. The change to poll for the status of the query is still an improvement to the general functionality of this adapter that should be considered.

### DIFF
--- a/lib/blazer/adapters/athena_adapter.rb
+++ b/lib/blazer/adapters/athena_adapter.rb
@@ -10,84 +10,30 @@ module Blazer
 
         begin
           resp =
-            client.start_query_execution(
-              query_string: statement,
-              # use token so we fetch cached results after query is run
-              client_request_token: Digest::MD5.hexdigest([statement,data_source.id].join("/")),
-              query_execution_context: {
-                database: database,
-              },
-              result_configuration: {
-                output_location: settings["output_location"]
-              }
-            )
+              client.start_query_execution(
+                  query_string: statement,
+                  # use token so we fetch cached results after query is run
+                  client_request_token: generate_client_request_token(statement),
+                  query_execution_context: {
+                      database: database,
+                  },
+                  result_configuration: {
+                      output_location: settings["output_location"]
+                  }
+              )
+
           query_execution_id = resp.query_execution_id
+          query_state = wait_for_query_results(query_execution_id)
 
-          timeout = data_source.timeout || 300
-          stop_at = Time.now + timeout
-          resp = nil
-
-          begin
-            resp = client.get_query_results(
-              query_execution_id: query_execution_id
-            )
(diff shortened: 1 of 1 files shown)


ankane:
Hey @bobby-vandiver, thanks for this as well. Since #256 already addresses the queued state, I don't think there's much benefit to using a separate endpoint to check the status (plus it adds an additional API call). However, if the current method breaks a few times, I'm happy to reconsider.

From the code comment on the request token change, it sounds like you're trying to have more control over caching. Does the idea discussed in #254 cover your use case? If not, can you explain your use case a bit more?

bobby-vandiver:
@ankane Sorry for not getting back to this sooner. 

Looking at #254, it seems the recommendation is to extend the existing adapter and handle the cache control at that point? For my use case it is sufficient to let the AWS SDK generate the `client_request_token` for every request. If I extend the existing adapter, I still have to duplicate the bulk of the logic for the adapter, which I want to avoid for maintainability. Separating the `client_request_token` into a separate method (`generate_client_request_token`) that can be overridden by a subclass (or outright monkey patched) accomplishes this goal for me. If that seems reasonable, I can create a separate PR with just that change.

As for the queued state in #256, I initially looked at a similar solution to this problem (the changes in this PR predate the resolution of #256 in our project), but I was not able to find any documented guarantee of the error messages coming from either AWS or the AWS SDK. The only reliable way to handle this that I could find was the approach I implemented in this PR to explicitly check the state of the query. Could you point me to where this is documented? I'm leery of depending on an undocumented error message in a production system.

Thanks!

ankane:
Hey @bobby-vandiver, thanks for the response. The existing adapters are essentially private APIs, so I don't think it makes sense to make changes to make them more extensible (since they could break at any time). You're better off writing your own custom adapter from scratch (which is a public API), which you could copy from an existing adapter and customize as needed.

That being said, if the change you want to make is generating a new request token with each request, there's a proposal for how to do that with `blazer.yml` configuration - see https://github.com/ankane/blazer/issues/254#issuecomment-520086546.

Re error message: there isn't a guarantee, but I can't recall it breaking in the past. If you'd prefer to use a documented API (which could be more reliable in the future), a custom adapter is probably the way to go.

ankane:
Cleaning up stale PRs

PR #281, opened 2020-03-22
Fix search by non-ASCII characters

aki77:
The `\W` character class in JavaScript RegEx is really only useful with simple ASCII encoding.

### DIFF
--- a/app/views/blazer/queries/home.html.erb
+++ b/app/views/blazer/queries/home.html.erb
@@ -71,7 +71,7 @@
   }
 
   var prepareQuery = function (str) {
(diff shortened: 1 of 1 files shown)


app/views/blazer/queries/home.html.erb
```diff
@@ -71,7 +71,7 @@
   }
 
   var prepareQuery = function (str) {
-    return str.toLowerCase().replace(/\W+/g, "")
```
aki77: It seems unnecessary because there is a fuzzy search.

ankane:
Looks great, thanks @aki77!

Issue #81, opened 2016-08-11
Add Materialized Views to the Preview Tables


strivedi183:
First off, thanks for a great library

One thing that I think would be useful would be to preview any materialized views in /queries/new in the Preview Tables drop down.

I'm not sure off the top of my head if all SQL databases support Materialized Views (we use Postgres) but for the ones that do, I think would be useful. If this is something you would be interested in having in this library, I'd be happy to help out with it anyway I can or create a PR and any suggestions on how you would go about it or at least a starting point would be appreciate.

Thanks again!

ankane:
Hey @strivedi183, I'd be happy for Blazer to have this feature. Here's where the [list of tables and schema](https://github.com/ankane/blazer/blob/master/lib/blazer/adapters/sql_adapter.rb#L43) are fetched. If you need to write PostgreSQL specific code, you can use the `postgresql?` method to conditionally add them.

ttilberg:
Some notes re: implementation for MS SQL Server for anyone tackling this in the future:

The functional equivalent to pg's materialized view in SQL Server is called an [Indexed View](https://msdn.microsoft.com/en-us/library/ms191432.aspx). It is created by first creating a view, and then adding a clustered index to it. This causes SQL Server to store the results of the view similar to storing a table.

There are a some requirements that must be met, and become imposed when you do this. Specifically some keywords are banned from the select query (like outer joins, count(*), min/max, and a variety of others). The view must also be created `WITH SCHEMABINDING` so the underlying tables cannot be altered if the schema change would affect the view, until the view itself is modified. I don't know if these types of restrictions are common on the other rdbs, so I wanted to explicitly note them.

From a strict tsql perspective it looks simple enough:

```sql
CREATE VIEW dbo.v_dragon_ravage_report WITH SCHEMABINDING
AS
  SELECT dragon, zip_code, sum(damage) as score, population
  FROM dragons d
  JOIN pillaged_vilages v ON d.id = v.dragon_id
  JOIN zip_codes z ON z.zip_code = v.zip_code
GO

CREATE UNIQUE CLUSTERED INDEX IDX_cl_v_dragon_ravage_report1
    ON dbo.v_dragon_ravage_report(dragons, zip_code)
GO
```

ankane:
Moving to ideas list in #24

PR #282, opened 2020-04-06
Add support for materialized views in the preview table dropdown when creating or editing a query.

strivedi183:
Add support for materialized views in the preview table dropdown when creating or editing a query.
This addition is only for PostgreSQL.

This PR is related to this issue #81 
It's not on the Ideas issue #24  so I'm not sure if you are still interested in adding this to Blazer.

Thanks again and this is a really great project!

### DIFF
--- a/lib/blazer/adapters/sql_adapter.rb
+++ b/lib/blazer/adapters/sql_adapter.rb
@@ -41,6 +41,7 @@ module Blazer
 
       def tables
         sql = add_schemas("SELECT table_schema, table_name FROM information_schema.tables")
+        sql = sql + " UNION ALL SELECT schemaname AS table_schema, matviewname AS table_name FROM pg_matviews" if postgresql?
         result = data_source.run_statement(sql, refresh_cache: true)
         if postgresql? || redshift? || snowflake?
           result.rows.sort_by { |r| [r[0] == default_schema ? "" : r[0], r[1]] }.map do |row|
--- a/lib/blazer/adapters/sql_adapter.rb
+++ b/lib/blazer/adapters/sql_adapter.rb
@@ -40,8 +40,29 @@ module Blazer
       end
 
       def tables
-        sql = add_schemas("SELECT table_schema, table_name FROM information_schema.tables")
-        sql = sql + " UNION ALL SELECT schemaname AS table_schema, matviewname AS table_name FROM pg_matviews" if postgresql?
+        sql =
+          if postgresql?
+            <<-SQL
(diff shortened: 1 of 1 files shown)


ankane:
Hey @strivedi183, thanks for the PR! Happy to include this feature, but I think it'll be a bit more work. The current query doesn't take into account:

1. If the Blazer user has privileges for the materialized views
2. The schema settings (applied by the `add_schemas` method)

I think the query below will cover the first one (based on [the query](https://github.com/postgres/postgres/blob/328c70997bc3518a50bd9a8ff33de349a7223413/src/backend/catalog/information_schema.sql#L1906-L1908) for the `information_schema.tables` view).

```sql
UNION ALL (
  SELECT n.nspname as table_schema, c.relname as table_name
  FROM pg_catalog.pg_class c
  LEFT JOIN pg_catalog.pg_namespace n ON n.oid = c.relnamespace
  WHERE c.relkind = 'm' AND
  (
    pg_has_role(c.relowner, 'USAGE')
    OR has_table_privilege(c.oid, 'SELECT, INSERT, UPDATE, DELETE, TRUNCATE, REFERENCES, TRIGGER')
    OR has_any_column_privilege(c.oid, 'SELECT, INSERT, UPDATE, REFERENCES')
  )
)
```

We'll need to add the materialized view columns to the `schema` method as well.

strivedi183:
Thank you for your feedback and help @ankane !

I tried to take into account both things you mentioned.

For 1), your query worked like a charm!

For 2), using the above query as a base, I used `pg_attribute.attname` to get the column_name, `pg_type.typname` to get the data_type, `pg_attribute.attnum` to get the ordinal_position.

It isn't an exact 1-1 match for the data_type naming, since `pg_type.typname` sometimes uses different names for the same type like `int4` and `int8` instead of `integer`. The types are still correct, just not consistent naming between tables and materialized views.

Thank you again!

strivedi183:
Hi @ankane , we've been running this in production for the past few months and we haven't had any issues.

After you get a chance to review, if there are any other issues, more than happy to fix them.

Thanks again!

notactuallytreyanastasio:
If there is anything left that would be necessary to do to make this merge-able? Would be a super useful feature for me so happy to contribute.

Aemrox:
Ditto on the above, this would be very useful

ankane:
Hi @strivedi183, thanks again for the PR, and sorry for the very long delay. Added support in the commit above.

PR #284, opened 2020-04-10
added gridstack so you can actually have 

dorkalev:
added gridstack to the dashboard view

ain't it nice?

image


ankane:
Hey @dorkalev, thanks for the PR, and sorry for the long delay (have been focused on other projects). It's a neat idea, but I'd like to keep the current dashboard format for now.

coorasse:
Hi @dorkalev this is a great idea! could you please rebase for me? 😄 
I'd like to copy-paste your work or use your branch directly. This is exactly what we were looking for!
@ankane what would be required in your opinion to have it integrated? Maybe that it can be enabled as an option? So that by default it keeps the one column layout.

PR #297, opened 2020-07-29
Added name validation to query & dashboard

pandu-49:
Added presence and uniqueness validation to queries and dashboards. Uniqueness validation score creator_id so that a user can have unique name which makes easy to identify and it avoids confusion.

### DIFF
--- a/app/models/blazer/dashboard.rb
+++ b/app/models/blazer/dashboard.rb
@@ -1,10 +1,11 @@
 module Blazer
   class Dashboard < Record
+
+    include ::Blazer::NameValidator
+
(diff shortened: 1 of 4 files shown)


ankane:
Hey @pandu-49, thanks for another PR! I like the idea. It should probably only run the uniqueness validation when the name changes, so existing queries and dashboards that have duplicate names won't be affected. Also, an empty string in the name of query indicates a private query, so it'll need to support that. Let's keep the validations inline for each of the models, even if there's a small bit of duplication.

pandu-49:
@ankane Thanks for the response. Could explain me more about private queries? Any example!

PR #298, opened 2020-08-01
Show the description on the Queries home page

bbugh:
Hello again open source hero ankane! 👋.

I was surprised to see that the queries index page does not show the query description. We have a number of queries that benefit from a few extra details that are provided in the description. This small PR adds the description to the query table for extra details.

For example:

image


ankane:
Hey @bbugh, thanks for the PR, and sorry for the delay (have been focused on other projects). I think descriptions will make the page a bit more cluttered, so I'd prefer to keep the current design.

PR #302, opened 2020-09-03
Fixed redirection error on deleting dashboard

pandu-49:
This will resolve the error that is raising while deleting dashboard because of URL not found error.


ankane:
Thanks @pandu-49! Looks like this was broken almost two years ago in 8c4f69dd5340c06b0782d1b9eecc930014be11af (huge facepalm).

Issue #270, opened 2019-11-20
[Idea] Limit setting for queries


VicenteFava:
Hi,

What do you think about having a limit setting that applies to all the queries if it is set? The idea of this is to prevent the database or the server from having performance issues when executing queries that grab a large amount of data.

If you think that this can be helpful I can do a PR with the code. Thanks!

ankane:
Hey @VicenteFava, thanks for the suggestion! For the database, it's best to set a statement timeout to limit the impact of individual queries. For the server, it'd be good to benchmark where most of the time is spent for large result sets to help decide what to do and test potential solutions. I think a `row_limit` option would likely help - we'd just need to make it clear to users when the results were limited (including for cached results).

joelvh:
Hi @ankane, we are using Blazer for some visualizations as well as custom reports. We'd also like to have something like a `row_limit` for the web view, to limit results which are displayed (or maybe we think of it akin to pagination), while still indicating the total row count.

I've looked at the code and there are several ins and outs that make it hard to know the best place to introduce this limit. We're just looking for a limit on what is displayed in a tabular view. Limiting data for charts isn't a focus for us at the moment. Just tabular view, with the ability to then download the full results.

Could you give some pointers of where best to look at implementing this? I'm happy to put a PR together.

Thanks!

rdlugosz:
A nice feature in addition to this would be to impose the row limits on the display version of the results, but allow for downloading the CSV of the full report results.

aki77:
#311 PR has already addressed limiting the number of html views only.

aki77:
Until the PRs are merged, I am using the following monkey patch.

_app/controllers/blazer/queries_controller.rb_
```ruby
require Blazer::Engine.root.join('app/controllers/blazer/queries_controller.rb')

module BlazerQueriesControllerExtention
  def render_run
    if request.format == :html && Blazer.settings['row_limit']
      @row_limit ||= Blazer.settings['row_limit']
    end

    super
  end
end

Blazer::QueriesController.prepend(BlazerQueriesControllerExtention)
```

_config/blazer.yml_
```yaml
row_limit: 5000
```

PR #311, opened 2020-11-16
Add row_limit settings

aki77:
I think this is useful in cases where the execution of a query doesn't time out, but the number of results is so large that it takes a long time to render.

I actually tested this with a simple query that returned 100,000 rows of results.
What would normally take more than 20 seconds to render was reduced to less than 5 seconds.

#270


aki77:
@ankane 

What do you think about this PR?
I'm having trouble with occasional server failures due to queries being executed with too many results.
Even though I have set the timeout to 10 seconds, it hasn't helped because the query is running so fast.

Issue #168, opened 2017-09-22
IN NEED OF 'Yesterday'


kdk2121:
Hi, our team is using blazer very well, and we think we need "Yesterday" in date range.
It is not only we look for data of yesterday much but since google analytics also support "Yesterday" as date range.

Think **ranges** in
**_variables.html.erb**
should be changed 
```
ranges: {
             “Today”: [dateStr(), dateStr()],
             “Last 7 Days”: [dateStr(6), dateStr()],
             “Last 30 Days”: [dateStr(29), dateStr()]
            },
```
Can you make changes plz?

JayTeeSF:
fwiw, I want "3 Days ago" ...it would be ideal, if that list of ranges (and the default) was configurable.

ankane:
Dup of #135

PR #313, opened 2020-11-19
Allow to configure preset time ranges

coorasse:
Everyone wants it different, I think we should provide a default, but also allow to change it.
With this PR you can now configure it freely.

image

Tackles #135 and #168

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 2.3.1 (unreleased)
 
 - Improved column names for uploads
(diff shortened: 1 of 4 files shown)


jnunemaker:
@ankane I would love this. Anything I can do to get it merged/released? Happy to put in the work as I use blazer on all my projects and this is my top wish. I'd be willing to be added to repo and maintain it going forward as well.

conarro:
@ankane anything we can do to get this one in? Happy to help, would love for this to be configurable

Issue #315, opened 2020-12-16
Problem with clear_helpers added in 2.4 and embedding Blazer within another app


denny:
Hi,

I'm loading Blazer inside the admin area of my application, so Blazer's templates define the main content area of the relevant pages, but my admin area's layout, menus, etc are rendering around the edges.

The change on line 9 here wipes out all of my app's helpers, which are used in those surrounding templates - meaning that none of my admin pages wrapping Blazer will load with Blazer 2.4:
https://github.com/ankane/blazer/blob/6048a2fdead7f361f1bd156334e7b7d2893d5346/app/controllers/blazer/base_controller.rb#L9

I wonder if it clearing the helpers could be made a configurable option? I have no opinion on the best default - I'm guessing my usage pattern is a minority one, and it sounds like this addition has fixed problems for other people - but it'd be great to have the option to keep my helpers if possible.

Thanks,
Denny

ankane:
Future readers, see #316

PR #316, opened 2020-12-17
Add config option to skip clear_helpers

denny:
Adds a config option to skip the clear_helpers call in the base controller; fixes my issue #315

### DIFF
--- a/app/controllers/blazer/base_controller.rb
+++ b/app/controllers/blazer/base_controller.rb
@@ -6,7 +6,7 @@ module Blazer
     skip_after_action(*filters, raise: false)
     skip_around_action(*filters, raise: false)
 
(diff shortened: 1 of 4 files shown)


ankane:
Hey @denny, thanks for the PR! Blazer views aren't designed to wrapped in application templates, so don't want to add an option for it, but you can probably add the helpers back manually if needed.

ankane:
Cleaning up PRs

denny:
In case anybody else does trip over this, here's what I have in my ApplicationController now:
```ruby
def self.clear_helpers
  super unless self == Blazer::BaseController
end
```
(Pretty sure I wouldn't want anything else to do it either, but I figured it was safest to override the minimum necessary for now.)

PR #321, opened 2021-01-14
Make naming of BaseController explicit

tbk303:
In my application I have a `BaseController` just like Blazer has. It seems that depending on non-deterministic loading order sometimes Blazer tries to inherit from my `BaseController` instead of the actual Blazer-`BaseController`.

This PR fixes this by making the inheritance of `Blazer::BaseController` explicit.

### DIFF
--- a/app/controllers/blazer/checks_controller.rb
+++ b/app/controllers/blazer/checks_controller.rb
@@ -1,5 +1,5 @@
 module Blazer
-  class ChecksController < BaseController
+  class ChecksController < Blazer::BaseController
     before_action :set_check, only: [:edit, :update, :destroy, :run]
 
     def index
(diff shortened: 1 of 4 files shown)


ankane:
Hey @tbk303, thanks for PR. However, I'm not sure how to reproduce the issue this addresses.

tbk303:
Just create a controller called `BaseController` in a Rails application where Blazer ist loaded (mounted in routes.rb).  I had issues with Blazer correctly inheriting from the `Blazer::BaseController`, instead it sometimes tried to inherit from my `BaseController`.

ankane:
Unfortunately, that doesn't reproduce it for me (and if it did, it'd likely be an issue with Rails autoloading rather than Blazer).

Issue #333, opened 2021-05-03
Blazer UUID problem


alec-c4:
Hi!
Blazer doesn't work with UUID identifiers. 

My migration:
```
class InstallBlazer < ActiveRecord::Migration[6.1]
  def change
    create_table :blazer_queries, id: :uuid do |t|
      t.references :creator, type: :uuid
      t.string :name
      t.text :description
      t.text :statement
      t.string :data_source
      t.string :status
      t.timestamps null: false
    end

    create_table :blazer_audits, id: :uuid do |t|
      t.references :user, type: :uuid
      t.references :query, type: :uuid
      t.text :statement
      t.string :data_source
      t.datetime :created_at
    end

    create_table :blazer_dashboards, id: :uuid do |t|
      t.references :creator, type: :uuid
      t.string :name
      t.timestamps null: false
    end

    create_table :blazer_dashboard_queries, id: :uuid do |t|
      t.references :dashboard, type: :uuid
      t.references :query, type: :uuid
      t.integer :position
      t.timestamps null: false
    end

    create_table :blazer_checks, id: :uuid do |t|
      t.references :creator, type: :uuid
      t.references :query, type: :uuid
      t.string :state
      t.string :schedule
      t.text :emails
      t.text :slack_channels
      t.string :check_type
      t.text :message
      t.datetime :last_run_at
      t.timestamps null: false
    end
  end
end
```

My exception:

```
server  | Started GET "/assets/blazer/favicon-392743107d0d14fea5699cff1e80d209db7058ef660e643ff9c9f07e990ac084.png" for ::1 at 2021-05-03 23:14:07 +0300
server  | Started POST "/blazer/queries" for ::1 at 2021-05-03 23:14:14 +0300
server  |   User Load (0.2ms)  SELECT "users".* FROM "users" WHERE "users"."id" = $1 ORDER BY "users"."created_at" ASC LIMIT $2  [["id", "149eedae-3b29-4874-87fb-9c40e49d0a08"], ["LIMIT", 1]]
server  |   Role Load (0.5ms)  SELECT "roles".* FROM "roles" INNER JOIN "users_roles" ON "roles"."id" = "users_roles"."role_id" WHERE "users_roles"."user_id" = $1 AND (((roles.name = 'admin') AND (roles.resource_type IS NULL) AND (roles.resource_id IS NULL)))  [["user_id", "149eedae-3b29-4874-87fb-9c40e49d0a08"]]
[...]

PR #338, opened 2021-05-05
Fix for Blazer UUID problem

alec-c4:
Fix for https://github.com/ankane/blazer/issues/333


alec-c4:
@ankane please test it and merge it to master

Issue #345, opened 2021-05-26
Has undeclared runtime dependencies


bf4:
Steps to reproduce

1. Install gem
2. Run migrations
3. `bin/rails zeitwerk:check` raises `NameError: uninitialized constant Blazer::ActionMailer`

Investigation:

Gemspec depends on railties (not rails meta-gem, yay!) but 
https://github.com/ankane/blazer/blob/a505849e893abef50c4128c2b3bfc9fae5ad2c86/app/mailers/blazer/check_mailer.rb (and the [slack notifier](https://github.com/ankane/blazer/blob/a505849e893abef50c4128c2b3bfc9fae5ad2c86/app/mailers/blazer/slack_notifier.rb) ) depend on ActionMailer

Proposal

1. Add gemspec dependency on `activerecord`, `activejob`, `actioncontroller`, `actionmailer` (and perhaps others. I didn't look too closely)
2. Consider making the actionmailer and activejob dependencies optional.

I'm up for making a PR

ankane:
Hey @bf4, thanks for bringing this up. `railties` already includes `actioncontroller` and `activejob` is already optional (only included when `async: true`), so I think `actionmailer` is the only issue. Instead of including it in `app`, I've moved it to `lib` with `autoload`, so it should only be included when needed and `rails zeitwerk:check` now passes. Let me know if you still run into issues.

bf4:
Awesome.  I'll try it out.

bf4:
Confirmed working as advertised.  Booting Rails works.  Explicitly referencing `Blazer::CheckMailer` raises

```
NameError: uninitialized constant Blazer::ActionMailer
Did you mean?  ActionView
from .bundle/vendor/ruby/2.7.0/gems/blazer-2.4.7/lib/blazer/check_mailer.rb:2:in `<module:Blazer>'
```

Awesome

PR #346, opened 2021-05-28
Declare runtime dependencies

bf4:
per https://github.com/ankane/blazer/issues/345

Just for discussion, I also made myself a minimal PR to disable the mailer when there's no action mailer present.  (Obviously, it would be better to allow configuring a real non-action-mailer class, but I didn't want to muddy the mvp case https://github.com/bf4/blazer/pull/1 )

### DIFF
--- a/blazer.gemspec
+++ b/blazer.gemspec
@@ -15,8 +15,12 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = ">= 2.4"
 
-  spec.add_dependency "railties", ">= 5"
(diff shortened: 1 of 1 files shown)


ankane:
Hey @bf4, thanks for the PR as well. I think the commit above should address #345 without needed to modify the gemspec.

bf4:
Sure.  Deciding whether to make required implicit dependencies explicit in a Rails gem is kind of a 'whatever feels right' decision.

PR #351, opened 2021-06-16
Stop establishing the connection as part of `preview_statement`

bnjix:
### Problem
`#preview_statement` is called in the `queries/_form` partial.

That has the side effect of needing to establish a connection for all sources that use the `SqlAdapter` as part of the `sqlserver?` method.

This causes the `queries/new` and `queries/edit` to become slow when one has a lot of data sources. 

### Solution
Use the url to determine if the connection is SQL Server or not

### DIFF
--- a/lib/blazer/adapters/sql_adapter.rb
+++ b/lib/blazer/adapters/sql_adapter.rb
@@ -208,7 +208,7 @@ module Blazer
       end
 
       def sqlserver?
-        ["SQLServer", "tinytds", "mssql"].include?(adapter_name)
+        data_source.settings["url"].starts_with?("jdbc:sqlserver://")
       end
 
       def snowflake?


ankane:
Hey @bnjix, thanks for the PR. However, it looks like the change doesn't cover all adapters. The connection should already be established on most page loads, so I'm not sure it'll make a big difference.

bnjix:
Hey @ankane thanks for the reply here!

It is true that this change does not cover all adapters. But in a case where one has most/all of its data sources being SQL this improves the load times significantly.

>The connection should already be established on most page loads

For `queries/new` and `queries/edit`, the only reason the connections are required is to know how to populate the preview statements for each DB, and that's why this change - by removing the need for the connection - greatly improves the performance of these 2 pages if one has many data sources.

ankane:
Hey @bnjix, thanks for the follow up. By not covering all adapters, I meant it replaces a check for 3 different SQL Server adapters (`SQLServer`, `tinytds`, `mssql`), with one (`jdbc:sqlserver://`), which will break existing installations.

PR #348, opened 2021-06-04
Adds viewport meta tag for dashboard

prithvi16:
[Viewport meta tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Viewport_meta_tag) sets viewport width to devise screen size. This gives a much better browsing experience for dashboard on mobile devices. 

Before on moto G4
Screenshot 2021-06-04 at 12 23 36 PM

After on moto G4
Screenshot 2021-06-04 at 12 24 07 PM

Before on ipad
Screenshot 2021-06-04 at 12 25 02 PM

After on ipad
Screenshot 2021-06-04 at 12 24 35 PM

before on iphone 5
Screenshot 2021-06-04 at 12 25 22 PM

After on iphone 5
Screenshot 2021-06-04 at 12 25 45 PM

PR #359, opened 2021-08-25
Make blazer responsive

dvisockas:
This makes blazer more mobile friendly 

Screenshot from 2021-08-25 20-55-00


ankane:
Hey @dvisockas, thanks for the PR. See #348.

PR #363, opened 2021-09-15
Skip basic auth if the environment variable `BLAZER_PASSWORD` is blank

coorasse:


ankane:
Hey @coorasse, thanks for the PR. However, I don't want to touch that logic right now.

PR #368, opened 2021-11-27
Fix multiple channel support for Slack apps

jeromecornet:
With the way slack apps work now, one of the limitations of the slack integration is that the webhook is specific to a single channel, so you need to configure slack apps with one webhook per channel you can post on.

Right now, specifying one or more channels in a check configuration does nothing with such an integration, because slack will post to the channel that's associated with the webhook, not the channel specified in the payload.

This PR tweaks the configuration so the slack_webhook_url config can either take a string (for the current behavior) or a mapping of slack channel -> webhook URL in form of a hash.

When sending the webhook, the new code will check whether a url is configured for this channel before sending it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,14 @@ BLAZER_SLACK_WEBHOOK_URL=https://hooks.slack.com/...
 
 Name the webhook “Blazer” and add a cool icon.
 
+If you want to send more than 1 channel, [create an app](https://api.slack.com/apps?new_app=1), create one webhook per channel
+you want to send to, and use the following syntax in blazer.yml
+```yaml
+slack_webhook_url:
+  a_channel: https://hooks.slack.com/...
+  another_channel: https://hooks.slack.com/...
+```   
+
 ## Authentication
 
 Don’t forget to protect the dashboard in production.
--- a/lib/blazer/slack_notifier.rb
+++ b/lib/blazer/slack_notifier.rb
@@ -49,7 +49,12 @@ module Blazer
         ]
       }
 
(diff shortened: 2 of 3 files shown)


ankane:
Hey @jeromecornet, thanks for the PR! If you use the legacy webhook integration (linked in the readme), it should be possible to post to any channel.

This approach looks pretty straightforward, but I'm hoping there's a way to set things up where channels don't need to be configured individually (still need to look into it).

jeromecornet:
Thanks @ankane . I haven't been able to configure an old-style integration with a recently created slack workspace, so this pr the best I could come up with to work around that. 

That said, if there is a simpler way I'm all for it 😊

ankane:
Hey @jeromecornet, just tried the legacy webhook integration with a new workspace and it still seems to work.

I also added a `slack_oauth_token` option that takes a bot token. It requires the `chat:write` and `chat:write.public` scopes, and for private channels, you need to invite the app with `/invite @blazer`. If you have some time, please try it out.

jeromecornet:
Thanks @ankane your approach is much cleaner especially because it avoids per-channel configuration which is much better.

For some reason I could not attach the legacy webhooks integration to my workspace or didn't understand the instructions as the webhook url I got was channel-specific and not workspace-specific.

That said, I tested your unreleased version on a an app with the oauth token by granting it the 2 permissions and it works well.

Thanks for looking into it and implementing it, I'll just close this PR as it's not the right approach

ankane:
Great, thanks for checking! Will probably make this the recommended approach in the next release.

PR #375, opened 2022-01-06
Check stv_inflight for query PID for redshift cancel

MasterOdin:
For some setups of Redshift, we are finding that getting the currently running query is more reliable to search the [`STV_INFLIGHT`](https://docs.aws.amazon.com/redshift/latest/dg/r_STV_INFLIGHT.html) over [`STV_RECENTS`](https://docs.aws.amazon.com/redshift/latest/dg/r_STV_RECENTS.html). In these setups, the query will show up in the former table and not in the latter table until after the query has finished. I'm not totally certain why that is, but suspect it's something to do with the nature of cluster nodes, but the whole thing is opaque and I couldn't find any good concrete information on googling.

The join here shouldn't add too much additional overhead, but I don't have a heavily used Redshift cluster to verify against.

### DIFF
--- a/lib/blazer/adapters/sql_adapter.rb
+++ b/lib/blazer/adapters/sql_adapter.rb
@@ -111,7 +111,8 @@ module Blazer
         if postgresql?
           select_all("SELECT pg_cancel_backend(pid) FROM pg_stat_activity WHERE pid <> pg_backend_pid() AND query LIKE ?", ["%,run_id:#{run_id}%"])
         elsif redshift?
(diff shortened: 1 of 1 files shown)


MasterOdin:
Not sure why the build suite is failing here, and I don't know enough about Ruby ecosystem or this project to know where to start.

ankane:
Hey @MasterOdin, thanks for the PR! Will dig into this next time I'm working with Redshift (not sure when that'll be exactly).

PR #376, opened 2022-01-11
Fix downloading of queries with a floating number

jplot:


ankane:
Hey @jplot, thanks for the PR. However, it appears to no longer export a CSV. Can you give more context on the change?

PR #377, opened 2022-01-17
Added support for ClickHouse

### LINKED ISSUE #378: [Idea] Support for ClickHouse
Hi!

It would be great to have adapter for [ClickHouse](https://clickhouse.com/) since its growing popularity among columnar databases.
I use ClickHouse database in my Rails project along with PostgreSQL. I also use Blazer for PostgreSQL and find it very useful! It would be handy to have single interface for both databases.

Communication with ClickHouse server can be implemented via HTTP interface (the easiest way — to take [this](https://github.com/shlima/click_house) library). By default `clickhouse-server` listens for HTTP on port 8123.

I got the courage to open a [PR](https://github.com/ankane/blazer/pull/377). Will be happy if you find it useful.

foxy-eyed:
[ClickHouse](https://clickhouse.com/) adapter implementation (resolves https://github.com/ankane/blazer/issues/378).

Requires `click_house` Ruby driver to interact with clickhouse-server via HTTP interface.

Adapter methods:
 - run_statement
 - tables
 - schema
 - preview_statement
 - explain
 
Readme instructions also added.

### DIFF
--- a/lib/blazer/adapters/clickhouse_adapter.rb
+++ b/lib/blazer/adapters/clickhouse_adapter.rb
@@ -1,14 +1,55 @@
 module Blazer
   module Adapters
     class ClickhouseAdapter < BaseAdapter
+      SUPPORTED_DRIVERS_MAPPING = {
+        "click_house" => ClickHouseDriver,
+        "clickhouse-activerecord" => ClickhouseActiverecordDriver
+      }.freeze
+
+      delegate :tables, to: :driver
+
+      # Wrapper for ClickHouse Ruby driver (https://github.com/shlima/click_house)
+      class ClickHouseDriver
+        delegate :tables, to: :connection
+
+        def initialize(config)
+          @config = ClickHouse::Config.new(**config)
+        end
+
+        def connection
+          @connection ||= ClickHouse::Connection.new(@config)
+        end
+
+        def execute(statement, format)
+          connection.post(query: { query: statement, default_format: format }).body
+        end
+      end
+
+      # Wrapper for Clickhouse::Activerecord driver (https://github.com/PNixx/clickhouse-activerecord)
+      class ClickhouseActiverecordDriver
+        delegate :tables, to: :connection
+
+        def initialize(config)
+          @config = config
+        end
+
+        def connection
+          @connection ||= ActiveRecord::Base.clickhouse_connection(@config)
+        end
+
+        def execute(statement, format)
+          body = connection.do_execute(statement, format: format)
+          format.in?(%w[CSV CSVWithNames]) ? CSV.parse(body) : body
+        end
+      end
+
       def run_statement(statement, _comment)
         columns = []
         rows = []
         error = nil
 
         begin
-          response = connection.post(query: { query: statement, default_format: "CSVWithNames" })
-          rows = response.body
+          rows = driver.execute(statement, "CSVWithNames")
           columns = rows.shift
         rescue => e
           error = e.message
@@ -17,23 +58,18 @@ module Blazer
         [columns, rows, error]
       end
 
-      def tables
-        connection.tables
-      end
-
       def schema
         statement = <<-SQL
           SELECT table, name, type
           FROM system.columns
-          WHERE database = '#{connection.config.database}'
+          WHERE database = '#{config[:database]}'
           ORDER BY table, position
         SQL
 
-        response = connection.post(query: { query: statement, default_format: "CSV" })
-        response.body
-                .group_by { |row| row[0] }
-                .transform_values { |columns| columns.map { |c| { name: c[1], data_type: c[2] } } }
-                .map { |table, columns| { schema: "public", table: table, columns: columns } }
+        rows = driver.execute(statement, "CSV")
+        rows.group_by { |row| row[0] }
+            .transform_values { |columns| columns.map { |c| { name: c[1], data_type: c[2] } } }
+            .map { |table, columns| { schema: "public", table: table, columns: columns } }
       end
 
       def preview_statement
@@ -41,19 +77,24 @@ module Blazer
       end
 
       def explain(statement)
-        connection.explain(statement)
+        driver.execute("EXPLAIN #{statement.gsub(/\A(\s*EXPLAIN)/io, '')}", "TSV")
       end
 
       protected
 
-      def connection
-        @connection ||= ClickHouse::Connection.new(config)
+      def driver
+        @driver ||= begin
+          driver = SUPPORTED_DRIVERS_MAPPING.keys.find { |driver| installed?(driver) }
+          raise Blazer::Error, "ClickHouse driver not installed!" unless driver
+
+          SUPPORTED_DRIVERS_MAPPING[driver].new(config)
+        end
       end
 
       def config
         @config ||= begin
           uri = URI.parse(settings["url"])
-          options = {
+          {
             scheme: uri.scheme,
             host: uri.host,
             port: uri.port,
@@ -61,9 +102,14 @@ module Blazer
             password: uri.password,
             database: uri.path.split("/").last
           }.compact
-
-          ClickHouse::Config.new(**options)
         end
+      end
+
+      def installed?(driver_name)
+        Gem::Specification.find_by_name(driver_name)
+        true
+      rescue Gem::LoadError
+        false
       end
     end
   end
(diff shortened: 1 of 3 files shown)


ankane:
Hey @foxy-eyed, thanks for the PR! Overall, looks great.

Have you tried [clickhouse-activerecord](https://github.com/PNixx/clickhouse-activerecord)? It looks like they both have ~60k downloads, and I think we could use the SQL adapter if it works.

foxy-eyed:
Hi @ankane, thank you for feedback!

I saw `clickhouse_activerecord` but I prefer not to choose it)

My understanding is, that ClickHouse is made for (and good for) processing big analytical data, like user activity, events etc into reports. In real-time.
ActiveRecord feels like unnecessary abstraction that does not fit scenarios because:
 - singe events are not as interesting (with all its numerous attributes) as chains of events and aggregated data; we read large number of rows, but only a small subset of columns;
 - in most cases we need some kind of aggregated report => we have to build custom SQL including subexpressions, functions, specific response format etc => dsl does not fit;
 - `#create` method mostly useless because data is inserted in large batches (otherwise it's inefficient);
 - we don't have relations;
 - usually data is not mutable after is is added to the DB.

Also support of different response format (https://clickhouse.com/docs/en/interfaces/formats/) would be handy, but `clickhouse_activerecord` processes only JSON. In other cases it returns response body as a string (btw, `clickhouse_activerecord` also uses HTTP interface under the hood but with `Net::Http` instead of `Faraday`).

But if someone already uses `clickhouse_activerecord` we can take it, right?)

I added detection of driver dependency and created classes for both drivers with common interface, blazer adapter now can work with both. Please take a look, what do you think? Should we keep it both?

ankane:
The main benefit of `clickhouse-activerecord` is we wouldn't need to maintain a separate adapter (assuming it works with `SqlAdapter`).

ankane:
It looks like queries work with `SqlAdapter`.

```yml
data_sources:
  my_source:
    url: clickhouse://playground:clickhouse@play-api.clickhouse.com:8443/datasets?ssl=true
```

However, `information_schema` isn't available until 21.11.2.2-stable (https://github.com/ClickHouse/ClickHouse/pull/28691), so the queries for tables and the schema would need to be updated to support earlier versions.

Also, it looks like neither library typecasts date and time columns, which is needed for charting, and `FixedString(N)` types aren't showing up properly (`SELECT * FROM visits_v1 LIMIT 10` in data source above).

With that additional info, I'm leaning towards `clickhouse-activerecord` still being a better fit for Blazer.

foxy-eyed:
> and FixedString(N) types aren't showing up properly (SELECT * FROM visits_v1 LIMIT 10 in data source above).

It looks like dataset issue, not ruby drivers.
I checked the same query with curl, with [Tabix](http://ui.tabix.io/#!/sql) and with RubyMine database tool (with official clickhouse jdbc driver) — same result.

```
$ curl "https://play-api.clickhouse.com:8443/?query=SELECT+ClientIP6+FROM+datasets.visits_v1+LIMIT+3;&user=playground&password=clickhouse&database=datasets"
{=�������bn�|I
{=�������bn�|I
{=�������bn�|I
```

> Also, it looks like neither library typecasts date and time columns, which is needed for charting

My bad, forgot to check charting.
I found out that `click_house` lib correctly typecasts query result in JSON format if one of [Clickhouse::Extend::ConnectionSelective](https://github.com/shlima/click_house/blob/master/lib/click_house/extend/connection_selective.rb) methods was called.
You can check the demo with fix:
https://clickhouse-blazer-demo.herokuapp.com/blazer/queries/1
https://clickhouse-blazer-demo.herokuapp.com/blazer/queries/2

Unfortunately, I couldn't make it work with `clickhouse_activerecord`.

Now it looks like `click_house` is better option.

foxy-eyed:
Now I am little confused)
I'll try to summarize:

1. `clickhouse_activerecord` + sql-adapter:
	   + no need to maintain a separate adapter
	   - growing complexity of sql-adapter (need to add some 'if's to support clickhouse < 21.11.2.2-stable)
	   - issues with date/time columns (I guess we could handle it on our own, but => even more complexity)
	   - force to install `clickhouse_activerecord` for Blazer even if `click_house` already used in project

2. `click_house` + separate clickhouse adapter
	   + keep the code separate and simple
	   + date/time columns + charts based on them works
	   - one more class to maintain
	   - force to install `click_house` for Blazer even if `clickhouse_activerecord` already used in project

3. Support both libs + separate clickhouse adapter
	   + no need to install extra gem for Blazer (there's a high probability that one of them already installed)
	   + separate adapter still keeps the code quite simple
	   - one more class to maintain
	   +/- ability of charting depends on the choice of lib.

@ankane what do you think?

ankane:
Let me think on this for a bit. I don't really want to support multiple adapters, so it'll likely be 1 or 2.

foxy-eyed:
@ankane hi!)
Have you come to any conclusion?

ankane:
Let's go with option 2 since date/time columns already work. A few other notes:

1. If `DateTime` objects are converted to `Time` in the adapter, they'll appear in `Blazer.time_zone` in the UI
2. In the `schema` method, think we can use `WHERE database = currentDatabase()`
3. In the `config` method, let's only remove the starting `/` from the path for the database, like [this adapter](https://github.com/ankane/blazer/blob/34c60e3af5c6ca6cabae3a8a5cbfb7bd2211fe71/lib/blazer/adapters/hive_adapter.rb#L38)
4. Would be good to support the `ssl_verify` option in the YAML config

foxy-eyed:
@ankane hi!

> If DateTime objects are converted to Time in the adapter, they'll appear in Blazer.time_zone in the UI

I'm not sure I got it right, but as I see none of ClickHouse types (`Date`, `DateTime`, `DateTime64`) are converted by `click_house` driver to ruby `Time` objects:

 * CH `Date` parses as Ruby `Date`
 * CH `DateTime`, `DateTime64` —  as Ruby `DateTime`

`DateTime#to_s` produces strings like `2022-02-06T01:10:11+00:00` (with original tz offset) in the UI.

All other comments — done ✅ 
Hope it is ok now)

ankane:
Looks great! For times, we'll want Blazer to display them in `Blazer.time_zone` to be consistent with other adapters (rather than the original tz offset), but to do that, they'll need to be converted to `Time` objects (`value.to_time`) when building `rows`.

lib/blazer/adapters/clickhouse_adapter.rb
```diff
@@ -0,0 +1,84 @@
...
+            ssl_verify: settings.fetch("ssl_verify", false)
+          }.compact
+          ClickHouse::Config.new(**options)
+        end
+      end
+
+      def convert_time_columns(row, date_time_columns)
+        time_values = row.slice(*date_time_columns).transform_values!(&:to_time)
+        row.merge(time_values)
+      rescue NoMethodError
```
foxy-eyed, line 79: We cannot be absolutely sure that we handle `DateTime` object here, because built-in types can be [overwritten](https://github.com/shlima/click_house#type-casting). If conversion to `Time` fails, we return initial object.

foxy-eyed (commented):
@ankane I believe I fixed last thing from your notes)

I was hoping to use custom type casting as described [here](https://github.com/shlima/click_house#type-casting) and overwrite default behaviour (parse `DateTime` columns as `Time`). But it can affect main app if it uses `ClickHouse` for business logic.
That is why I came to a decision to process rows in adapter and apply conversion `DateTime#to_time` after type casting.

foxy-eyed:
Hi @ankane, sorry for bothering you. Do you have any other concerns about this PR? Can it be merged?

okdas:
Hi @ankane, this adapter would be extremely useful as clickhouse is a very popular column store often utilized as analytics data store. Are there any blockers for this pr? Thanks.

khasinski:
@abuisman I've forked the project and added Clickhouse support from this branch (including a small bugfix if the schema doesn't have any tables).

My plan is to selectively merge some PRs from blazer and rewrite a couple components to make it easier extending the project (for example, I'm using bootstrap 5 and stimulus for interactions). It's still a work in progress, but feel free to share your thoughts, [here is the fork](https://github.com/finery-bi/finery)

ankane:
Hi @foxy-eyed, thanks again for the PR, and sorry for the very long delay. Added support without an external library in the commit above.

PR #382, opened 2022-02-20
Add ability to set athena aws region for a single data source

cianmce:
We use Athena in multiple regions and don't want it globally set to a single region
E.g. currently we're using `ENV['AWS_REGION'] = "us-east-1"` but we want to have the ability to not have it globally set for the application and to have different `data_sources` with different regions


cianmce:
@ankane any feedback on this?

ankane:
Looks great, thanks @cianmce!

PR #383, opened 2022-02-20
[Mapbox] Prevent truncation in the middle of a HTML tag + add logic for better truncation based on key and value length

cianmce:
## Fixes/improvements
 - As in below screenshot, when the length of the content is such that there's a HTML `<br />` or `<strong>` tag at the 140 character position in the string, then the tag ends up being truncated to result in opening angle brackets appearing
 - IMO I don't think `<strong>` or `<br />` tags should count towards the char limit
 - previously a large column name could result in the content being needlessly truncated when the column can be instead
 - if there's 1 long row, that now won't prevent the proceeding columns being shown
   - now the earlier long row will instead be truncated

Eg
```sql
SELECT
  2.2913515 AS longitude,
  48.85391 AS latitude,
  "Eiffel Tower" AS name,
  "Paris" AS city,
  "France" AS country,
  "French" AS language_name,
  "300 m, 324 m to tip" AS height,
  "The Eiffel Tower is a wrought-iron lattice tower on the Champ de Mars in Paris, France. It is named after the engineer Gustave Eiffel, whose company designed and built the tower." AS full_wiki_description
```

## Before:

image

## After:

image

### With 1 long row

image

These limits could be made configurable later on if there was a need but I think these are reasonable limits

### DIFF
--- a/app/controllers/blazer/queries_controller.rb
+++ b/app/controllers/blazer/queries_controller.rb
@@ -264,10 +264,23 @@ module Blazer
               @rows.select do |r|
                 r[lat_index] && r[lon_index]
               end.map do |r|
+                rows_limit = 6
+                row_length_limit = 70
+                key_length_limit = 20
+                limit = 140
+                title = @columns.zip(r).reject { |k, v| keys.include?(k) }[0...rows_limit].collect do |k, v|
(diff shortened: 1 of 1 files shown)


ankane:
Hi @cianmce, thanks for the PR, and sorry for the delay. Just pushed a fix for the tag issue, but don't want to make any other changes right now.

Issue #364, opened 2021-04-14
Set character encoding


Enormity:
Having issues exporting with special characters like Chinese characters.  Would be great to be able to either set the character encoding in a config file or on the export level.

PR #386, opened 2022-02-23
Enable adapter tests and disable broken Elasticsearch tests

cianmce:
- include tests by updating pattern
 - add MySQL to be included during tests
 - skip broken Elasticsearch tests

I also tried adding [ankane/setup-elasticsearch](https://github.com/ankane/setup-elasticsearch) but that failed with the following error:

```ruby
The client is unable to verify that the server is Elasticsearch. Some functionality may not be compatible if the server is running an unsupported product.
The client is unable to verify that the server is Elasticsearch. Some functionality may not be compatible if the server is running an unsupported product.
.......E

Error:
ElasticsearchTest#test_tables:
Elasticsearch::UnsupportedProductError: The client noticed that the server is not Elasticsearch and we do not support this unknown product.
    /home/runner/work/blazer/blazer/vendor/bundle/ruby/3.1.0/gems/elasticsearch-8.0.0/lib/elasticsearch.rb:1[15](https://github.com/cianmce/blazer/runs/5305254255?check_suite_focus=true#step:7:15):in `verify_with_version_or_header'
    /home/runner/work/blazer/blazer/vendor/bundle/ruby/3.1.0/gems/elasticsearch-8.0.0/lib/elasticsearch.rb:104:in `verify_elasticsearch'
    /home/runner/work/blazer/blazer/vendor/bundle/ruby/3.1.0/gems/elasticsearch-8.0.0/lib/elasticsearch.rb:70:in `method_missing'
    /home/runner/work/blazer/blazer/vendor/bundle/ruby/3.1.0/gems/elasticsearch-api-8.0.0/lib/elasticsearch/api/namespace/common.rb:38:in `perform_request'
    /home/runner/work/blazer/blazer/vendor/bundle/ruby/3.1.0/gems/elasticsearch-api-8.0.0/lib/elasticsearch/api/actions/cat/indices.rb:60:in `indices'
    /home/runner/work/blazer/blazer/lib/blazer/adapters/elasticsearch_adapter.rb:30:in `tables'
    /opt/hostedtoolcache/Ruby/3.1.1/x64/lib/ruby/3.1.0/forwardable.rb:238:in `tables'
    /home/runner/work/blazer/blazer/app/controllers/blazer/queries_controller.rb:[20](https://github.com/cianmce/blazer/runs/5305254255?check_suite_focus=true#step:7:20)0:in `tables'
[...]

Issue #385, opened 2022-02-23
Encoding error dealing with binary data


cianmce:
We have been using Blazer with some tables that have binary UUID columns.
When returning data that has binary data and plain text data it raises an error.

e.g. running the following SQL with a MySQL data source

```sql
select unhex("F6"), 123
```

raises

```ruby
Encoding::UndefinedConversionError - "\xF6" from ASCII-8BIT to UTF-8:
```

This only happens if
 1. the binary column is the first column returned
 2. there's no non-binary columns after the binary columns

e.g. these all work
```sql
select unhex("F6"); /* this works */
select unhex("F6"), unhex("F6"); /* this works */
select 123, unhex("F6"); /* this works */
select 123, unhex("F6"), 123; /* this works */
```

We have found this monkey patch fixes the issue but it's not ideal.
Adding this to `config/initializers/blazer.rb`
```ruby
module SwrveBlazer
  module BaseHelper
    def blazer_format_value(key, value)
      format_value = super(key, value)
      format_value.respond_to?(:force_encoding) ? format_value.force_encoding(Encoding::UTF_8) : format_value
    end
  end
end

Blazer::BaseHelper.prepend(SwrveBlazer::BaseHelper)
```

cianmce:
This possibly also fixes
https://github.com/ankane/blazer/issues/364

PR #387, opened 2022-02-23
[385] fix binary data encoding issue with tests

cianmce:
Following on from issue: https://github.com/ankane/blazer/issues/385

Possibly also fixes https://github.com/ankane/blazer/issues/364

 - Builds on PR: https://github.com/ankane/blazer/pull/386 where MySQL tests are enabled
 - Force encoding of data to UTF-8
 - Added tests for binary encoded data

### Issue

We have been using Blazer with some tables that have binary UUID columns.
When returning data that has binary data and plain text data it raises an error.

e.g. running the following SQL with a MySQL data source

```sql
select unhex("F6"), 123
```

raises

```ruby
Encoding::UndefinedConversionError - "\xF6" from ASCII-8BIT to UTF-8:
```

This only happens if
 1. the binary column is the first column returned
 2. there's no non-binary columns after the binary columns

e.g. these all work
```sql
select unhex("F6"); /* this works */
select unhex("F6"), unhex("F6"); /* this works */
select 123, unhex("F6"); /* this works */
select 123, unhex("F6"), 123; /* this works */
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,4 +15,5 @@ jobs:
     - uses: ankane/setup-mysql@v1
       with:
         database: blazer_test
+    - uses: ankane/setup-elasticsearch@v1
     - run: bundle exec rake test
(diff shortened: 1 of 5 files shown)


ankane:
Hey @cianmce, thanks for reporting and the PR! It looks like it was due to the charting logic. Should be fixed in the commit above, but let me know if there are other queries that don't work.

Issue #388, opened 2022-02-23
[Idea] Option for applying special formatting to certain columns, e.g., Pretty-printing JSON


rdlugosz:
Hi - Similar to the concept of smart_columns, or perhaps as an extension of them, how about a way to use a block of Ruby code to format the value of a column when displaying on the page?

The specific use case I have in mind for this is that I've got a result set that includes some JSON in one of the columns. It would be nice if there was a way I could apply some pretty printing (line breaks, indenting, etc) to that value before it's displayed in the table of results.

I suppose this would open up quite a few other options for customization, similar to how you can pass a block to a field in ActiveAdmin and do whatever you'd like with the value before it's displayed (e.g., adding links, styles, etc).

Thanks for considering!

cianmce:
Something similar to what is done with images could be done here [app/helpers/blazer/base_helper.rb#L19](https://github.com/ankane/blazer/blob/a505849e893abef50c4128c2b3bfc9fae5ad2c86/app/helpers/blazer/base_helper.rb#L19) 
E.g. if it ends in `json` or `yaml` then attempt to pretty print and put it in a `<pre>` tag

PR #389, opened 2022-02-24
Add basic JSON and YAML rendering in pre tag

cianmce:
Following on from idea: https://github.com/ankane/blazer/issues/388

If a column name contains `JSON` or `YAML` then do the appropriate parsing and pretty-printing, if parsing fails then just return the value in a `pre` tag

We could also add config to explicitly enable or disable this, similar to what is done with image

<hr>

Sample SQL using MySQL data source
```sql
SELECT "test: [1,2,3,4]\ntest2: key2" AS sample_yaml, '{"test": [1,2,3,4], "test2": "key2"}' AS sample_json, "test_content" AS sample_text
UNION
SELECT "invaid: yaml\nobject", '{"invalid": "json"', "more_text"
```

image

### DIFF
--- a/app/helpers/blazer/base_helper.rb
+++ b/app/helpers/blazer/base_helper.rb
@@ -24,7 +24,21 @@ module Blazer
           link_to value, value, target: "_blank"
         end
       else
-        value
+        if key.include?("json")
+          begin
+            content_tag(:pre, JSON.pretty_generate(JSON.parse(value)))
+          rescue JSON::ParserError
+            content_tag(:pre, value)
+          end
+        elsif key.include?("yaml")
+          begin
+            content_tag(:pre, YAML.dump(YAML.load(value)))
+          rescue Psych::SyntaxError
+            content_tag(:pre, value)
+          end
+        else
+          value
+        end
       end
     end
 


jpaulomotta:
I think this is a great idea and it would be useful on projects that I'm working on.

cianmce:
@ankane is this something you'd look at including?

app/helpers/blazer/base_helper.rb
```diff
@@ -39,5 +45,23 @@ def blazer_js_var(name, value)
     def blazer_series_name(k)
       k.nil? ? "null" : k.to_s
     end
+
+    private
+
+    def format_json_value(value)
+      begin
```
alexshapalov, line 52: The use of begin...rescue inside methods is redundant in Ruby since you can directly use rescue

```
def format_json_value(value)
  JSON.pretty_generate(JSON.parse(value))
rescue JSON::ParserError
  value
end

def format_yaml_value(value)
  YAML.dump(YAML.load(value))
rescue StandardError
  value
end
```

app/helpers/blazer/base_helper.rb
```diff
@@ -39,5 +45,23 @@ def blazer_js_var(name, value)
...
+        JSON.pretty_generate(JSON.parse(value))
+      rescue JSON::ParserError
+        value
+      end
+    end
+
+    def format_yaml_value(value)
+      begin
+        YAML.dump(YAML.load(value))
+      rescue Exception
```
alexshapalov, line 62: It's safer to use the StandardError class instead of Exception.

app/helpers/blazer/base_helper.rb
cianmce: Good point, done

PR #393, opened 2022-04-21
Fixed undefined "prepared_statements?" method for Rails < 7.0

giemsky:
At the beginning I would like to thank you for this gem and all your work.

Looks like changes introduced in Blazer 2.6 has broken apps using MySQL and Rails < 7.0. 

The following exception is raised:
```
NoMethodError - undefined method `prepared_statements?' for #<ActiveRecord::ConnectionAdapters::Mysql2Adapter:0x000000010e905b78>
```

From what I researched `prepared_statements?` was added as an alias to `prepared_statements` in Rails 7 https://github.com/rails/rails/commit/da26910fbcfa3aa23ebc1474b43b048b78b60e4c

### DIFF
--- a/lib/blazer/adapters/sql_adapter.rb
+++ b/lib/blazer/adapters/sql_adapter.rb
@@ -193,7 +193,7 @@ module Blazer
       def parameter_binding
         if postgresql? || sqlite?
           :numeric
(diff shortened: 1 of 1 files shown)


ankane:
Thanks @giemsky! Will push out a new release shortly.

PR #402, opened 2022-06-10
Add temporal line chart annotations

### LINKED ISSUE #403: [Idea] Add time graphs annotations
Use chart-js annotations to overlay data on multiple time graphs at the same time.

This allows to quickly correlate events (both points in time and time ranges) with multiple graphs at the same time. Especially useful in dashboards. I've used this feature heavily in a clients code to see the influence of deployments and outages on business metrics.

Here is a demo screenshot:

image

And here is a PR: https://github.com/ankane/blazer/pull/402

khasinski:
I'd like to propose an annotations mechanism which allows overlaying data on top of the line charts with a temporal axis.

The main use case is to keep a log of events (whether they are points in time or ranges) and to see the effect of those events on multiple metrics we track in Blazer. A good example would be to track deployments, a marketing campaign duration or holidays and overlay them on top of revenue or visits data.

Here is an example:

blazer.yml
```yml
    annotations:
      deployments:
        SELECT created_at, name FROM deployments # datetime + string, results in a vertical line + label
      holidays:
        SELECT min_date, max_date, name FROM holidays # two dates + string, results in a colored range box
 ```

Result
image

Developed together with @twratajczak

Fixes #403

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,6 +266,32 @@ smart_columns:
   status: {0: "Active", 1: "Archived"}
 ```
 
+### Annotations
+
+Shows overlay lines or box ranges for line queries.
+
+Suppose your sales data and your deployments data, given a query:
+
+```sql
+SELECT date_trunc('hour', created_at), sum(value) FROM sales GROUP BY 1
+```
+
+You might want to see the influence of a deployment for those sales.
+
+```yml
+annotations:
+  deployments: SELECT date, name FROM deployments WHERE date BETWEEN {min_date} AND {max_date}
+```
+
+You can also show periods:
+
+```yml
+annotations:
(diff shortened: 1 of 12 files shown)


khasinski:
I've added the ability to override colors, even though it requires extending `Blazer::Annotations` and setting it with `Blazer.annotations`. This might be used to easily mark for example weekends with a light gray. Helpful if the data has a cyclical nature.

ankane:
Hi @khasinski, thanks for the PR, and sorry for the long delay. It's an interesting idea, but not something I'd like to support.

khasinski:
No problem, thanks for getting back to me with this!

PR #410, opened 2022-08-07
Add sharing feature

abuisman:
We are looking at replacing, amongst other things, Heroku dataclips with Blazer. The thing that we use a lot with dataclips is the sharing feature: it allows a developer to write a SQL query, they then share the CSV url with a non-technical person who can then import that into whatever tool that supports CSVs. Every time they re-download the CSV it is up-to-date with the query. When importing into Google sheets it results in an always up-to-date sheet, great all around.

In this PR I have added the feature to Blazer. 

image

In building I realised I can still extract it and sort of patch it in, but I'd rather contribute back. I've tried to keep close to the overall design and style.

The only change I needed to make to the `QueriesController` that might be a good change regardless of this feature is the following:

From:

```ruby
@statement = Blazer::Statement.new(params[:statement], @data_source)
```

To:

```ruby
sql_statement = params[:statement] || @query.statement
@statement = Blazer::Statement.new(sql_statement, @data_source)
```

### DIFF
--- a/app/controllers/blazer/queries_controller.rb
+++ b/app/controllers/blazer/queries_controller.rb
@@ -78,6 +78,14 @@ module Blazer
     def edit
     end
 
+    def share
+      if params[:token] && params[:query_id] && params[:token] == Blazer.sharing.query_token(params[:query_id])
+        run
+      else
+        render_forbidden
+      end
+    end
+
     def run
       @query = Query.find_by(id: params[:query_id]) if params[:query_id]
 
@@ -87,7 +95,8 @@ module Blazer
       data_source ||= params[:data_source]
       @data_source = Blazer.data_sources[data_source]
 
-      @statement = Blazer::Statement.new(params[:statement], @data_source)
+      sql_statement = params[:statement] || @query.statement
+      @statement = Blazer::Statement.new(sql_statement, @data_source)
       # before process_vars
       @cohort_analysis = @statement.cohort_analysis?
 
--- a/lib/blazer/sharing.rb
+++ b/lib/blazer/sharing.rb
@@ -1,0 +1,36 @@
+module Blazer
+  class Sharing
+    attr_accessor :api_key, :path
+
+    def initialize(api_key: ENV.fetch('BLAZER_DOWNLOAD_API_KEY', nil), path: '/blazer_share')
+      @api_key = api_key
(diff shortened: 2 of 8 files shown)


app/controllers/blazer/queries_controller.rb
```diff
@@ -78,6 +78,14 @@ def show
     def edit
     end
 
+    def share
+      if params[:token] && params[:query_id] && params[:token] == Blazer.sharing.query_token(params[:query_id])
```
bdewater, line 82: This can leak the token via timing attacks, so compare in constant time:
```suggestion
      if params[:token] && params[:query_id] && ActiveSupport::SecurityUtils.secure_compare(params[:token], Blazer.sharing.query_token(params[:query_id]))
```

lib/blazer/sharing.rb
```diff
@@ -0,0 +1,36 @@
...
+    def route_path
+      @route_path ||= "#{path}/:token/:query_id"
+    end
+
+    def to_controller
+      'blazer/queries#share'
+    end
+
+    def query_token(query_id)
+      Digest::SHA1.hexdigest("#{query_id}-#{api_key}")
```
bdewater, line 19: This should use an HMAC, which was specifically designed to address [the security issues with hashing simple concatenations of the key and message](https://en.wikipedia.org/wiki/HMAC#Design_principles). While we're at it we can use SHA-256 (future proof security) with base64 encoding (trim back the longer length compared with SHA-1).

```suggestion
      hmac = OpenSSL::HMAC.digest("sha256", api_key, query_id)
      Base64.urlsafe_encode64(hmac)
```

bdewater: IMO better alternative would be to add a column to the queries table that stores a randomly generated token, so revoking is possible on a per-query basis instead of all queries (by changing the API key). Could be implemented using [ActiveRecord::SecureToken](https://api.rubyonrails.org/classes/ActiveRecord/SecureToken/ClassMethods.html).

abuisman: I completely agree with your solution to use SecureToken instead of an API key. It is a lot better! 

I've implemented it in the latest commits. Thanks for suggesting it.

abuisman:
I've updated the design as suggested by @bdewater. I now generate a token per query which allows for easy per-query revoking of access. Great suggestion.

Issue #414, opened 2022-09-19
[Idea] Add internationalization


tvdeyen:
This gem is awesome. Thanks for releasing it.

It would be great if the interface could be internationalized (localized and translated), though.

Since this is a Rails engine the `i18n` gem and all its features as well as the action view (`t`) helper are already present.

What we would need to do is:

1. Replace every user facing English string with a call to `I18n.t` in controllers/models or simply `t` in views with a proper key and to avoid conflicts the `blazer` `scope`.
Ie. Instead of `"Select a date"` in `app/views/blazer/_variables.html.erb` we would use `<%= t(:placeholder, scope: [:blazer, :date_select]) %>`
2. Add the current strings as default locale into `config/locales/blazer.en.yml`
3. Add localizations for the timepicker used and configure it with current `I18n.locale`
4. (Optional) add a locale select into the admin UI

If this is something you would accept as a contribution I would start working on it and open a WIP PR to further discuss how to implement this.

tvdeyen:
@ankane are you interested in this feature? I would like to continue working on it, because I think it would be beneficial for lot's of people. WDYT?

ankane:
Future readers, see https://github.com/ankane/blazer/pull/415#issuecomment-1428772430

PR #415, opened 2022-09-19
Add internationalization

tvdeyen:
Refs #414


ankane:
Hey @tvdeyen, thanks for the suggestion and PR. I don't think it's something I can commit to supporting right now, but may revisit in the future.

PR #417, opened 2022-09-23
Add support for displaying GeoJSON columns in a map

kmarsh:
This adds support for rendering columns containing GeoJSON in a map. This works especially well with PostGIS where you can generate GeoJSON from geometry columns with `ST_AsGeoJSON`. e.g.,

```sql
SELECT id, name, ST_AsGeoJSON(the_geom) AS the_geom_geojson FROM table
```

image

This works great for visualizing data in tables that contain geographic data, but also for experimenting with and viewing queries with arbitrary data and/or geographic calculations. e.g.,

```sql
SELECT id, name, ST_AsGeoJSON(ST_Envelope(the_geom::geometry)) AS the_geom_geojson
FROM counties
WHERE state_abbreviation = 'OH'
  AND name ILIKE 'C%';
```

image

 The implementation was derived from markers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -354,10 +354,14 @@ SELECT gender, COUNT(*) AS pie FROM users GROUP BY 1
 
 ### Maps
 
-Columns named `latitude` and `longitude` or `lat` and `lon` or `lat` and `lng` - [Example](https://blazer.dokkuapp.com/queries/15-map)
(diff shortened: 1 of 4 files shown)


ankane:
Hey @kmarsh, thanks for the PR (and sorry for the delay)! Love this idea (and great choice of state for the example).

Blazer 3 (the next release) will use mapbox-gl-js / Mapkick.js for maps, so updated this to use it. Also, think it's slightly better to only match on the exact name, so made that change as well.

Let me know if you have a chance to try it and have any feedback.

PR #421, opened 2022-10-26
Making pageSize and setLimit constants equal

gotoAndBliss:
I'm not sure if pagination is intended to work here, but in its absence, the pageSize should equal the setLimit constant so that all items are shown.


ankane:
Hey @gotoAndBliss, thanks for the PR, but they shouldn't need to be equal.

PR #431, opened 2023-03-13
Additional info about smart variables

danmaz74:
It wasn't immediately obvious to me where the smart variable should be defined; with this minimal addition it should be more obvious :)


ankane:
Thanks @danmaz74!

PR #439, opened 2023-04-27
Add .table-responsive to .results-container

phuwanart:
Add table-responsive css class for results container.


jdelStrother:
I wish Blazer made better use of horizontal space for the results table... my results often end up looking something like this 

image

it seems a shame to constrain the table to a max width of 1140px with `table-responsive` if you do have a bigger screen ... but haven't managed to come up with anything that's more aesthetically appealing for the general case.

PR #440, opened 2023-06-14
Better description

mtrejo0:
When making a blazer the new lines are not preserved in the description making it hard to understand the goal of the blazer

If I have more than 4 lines of description it is all collapsed to 2 lines in a row. 

Most people skip this

Having it spaced out like this github description is one solution that will make it much clearer for people what a blazer does


ankane:
Hi @mtrejo0, thanks for the PR. Newlines should be preserved after 5c2d21f776e7292d73fdceb8ca444ca4dac55b8b (but it's not released yet).

PR #447, opened 2023-10-10
Allow role-based authentication for Athena

jdelStrother:
WDYT to something like this?

You can now specify role_arn to assume a role for querying Athena, rather than using the access key & secret directly.

~STS tokens expire after 1 hour by default, so I've also wrapped the client calls in a `autorefresh_credentials` block. (Which is a little gross, open to better suggestions)~
[I've since updated to using Aws::AssumeRoleCredentials, which handles autorefreshing for you]

STS requires you to specify a region, which might not be the same as the database's region. To try & clean up the distinction I've moved all the Athena credential settings to a new `credentials` sub-hash, but it should continue to work with people still setting access key & secret at the top level.


jdelStrother:
Let me know if this is something you might be interested in and I'll rebase. We've been monkey-patching it in our app and it's been working well for us

PR #448, opened 2023-10-13
CI: Tell dependabot to update GH Actions versions

olleolleolle:


ankane:
Hi @olleolleolle, thanks for the PR, but not interested in this.

PR #452, opened 2023-10-25
Fixed error with Sqlite when prepared statements are disabled with Rails < 6.1

dhnaranjo:


ankane:
Hi @dhnaranjo, thanks for reporting and the PR. Fixed in fe963b08cf0f6244f5fd18a3407fca5ffa276fbe (it still ignores them in Active Record 6.1+).

dhnaranjo:
Yup I did just discover that, forgot I had disabled the query logging before I tested it.

Always appreciate how on top of fixes you are. Dang.

Issue #279, opened 2020-03-20
Prepend UTF-8 BOM to CSV


kzkn:
I want to be able to prepend UTF-8 BOM (`"\xEF\xBB\xBF"`) to the downloaded CSV file.

MS Excel does not correctly recognize UTF-8 (without BOM, it's plain) files containing non-ASCII characters. If you prepend UTF-8 BOM, it will recognize correctly. I think this is a bug of MS Excel, but there is no way to fix it. We have no choice but to avoid it.

I have confirmed that it can be avoided by applying a monkey patch to `Blazer::QueriesController`.

```ruby
load Blazer::Engine.root.join('app/controllers/blazer/queries_controller.rb')
Blazer::QueriesController.prepend(Module.new do
  def csv_data(columns, rows, data_source)
    "\xEF\xBB\xBF" + super
  end
end)
```
SEE: https://stackoverflow.com/questions/2881093/override-a-rails-engine-controller-action

I want to take this as a function.
In my idea, it is good that it can be contorlled with `blazer.yml` to keep compatibility.

ankane:
Hey @kzkn, thanks for the suggestion. Which versions of MS Excel are affected?

kzkn:
@ankane 
As far as I know, version 2013 and earlier are affected at least.

kzkn:
@ankane 
If I send a pull request, could you merge it?

ankane:
Hey @kzkn, this isn't something I want to include in the gem right now. Thanks for sharing your approach. Others who need this should be able to use it.

abrambailey:
Agree this would be a nice option. We ran into the same issue with our project.

PR #454, opened 2023-11-07
Add `blazer_base_controller` load hook

aki77:
Added hook for easy customization of controllers.

Related Issues: #279


ankane:
Hi @aki77, thanks for the PR, but this isn't something I'd like to officially support.

PR #461, opened 2024-01-12
Cohorts: Remove zeros, show appropriate number of columns

tgig:
This change removes zeros from table so the cohort triangle can be seen easier. Also calculates the number of columns to display instead of hard-coding 12 of them.

Old:
Screenshot 2024-01-12 at 1 03 34 PM

New:
Screenshot 2024-01-12 at 1 03 59 PM


ankane:
Hi @tgig, thanks for the PR. Future cohorts already won't have zeros, and I'd like to keep the number fixed.

PR #473, opened 2024-05-23
Allow customizing parent controller of Blazer::BaseController

grk:
I ran into the problem of having admin user auth methods defined on a controller that inherits from ApplicationController, and thus being unable to properly secure blazer.

This PR (or more like a draft to start a discussion) adds an ability to customize that. 

By default it will behave as before, inheriting from ApplicationController, but it's customizable in `Settings['base_controller']`.


n-studio:
Great feature and very simple implementation. @ankane Is there any blocker?

PR #474, opened 2024-05-23
Changes download to not write file before sending response

ronualdo:
I have a few cases where the download action is triggering server timeouts because it seems that Blazer is writing the whole CSV to the disk before sending the response. I was wondering if we could use `response.stream` instead to avoid this type of scenario.

I've opened this PR with an initial implementation. Would love to hear any thoughts on this.


ankane:
Hi @ronualdo, thanks for the suggestion and PR! I think we can switch to `send_stream` once Rails 6.1 is no longer supported ([branch](https://github.com/ankane/blazer/compare/stream-csv)).

shahjapan (approved):
Looks good to me.

Issue #481, opened 2024-09-04
Frozen array error with Trilogy and Rails 7.2


jenshenny:
When using the Trilogy adapter and Rails 7.2, I'm getting a frozen array error for every query. 

<img width="208" alt="Screenshot 2024-09-04 at 11 54 28 AM" src="https://github.com/user-attachments/assets/28958530-09c9-475d-993e-d1cf0dfebb15">

After looking into it, `columns` is now frozen in Rails 7.2. https://github.com/rails/rails/commit/9652207c6b12ccdfd7067b874f8ceff10c54d2b4

To fix it we can dup the columns array on https://github.com/ankane/blazer/blob/master/lib/blazer/adapters/sql_adapter.rb#L47

```
columns = columns.map { |k| k.dup.force_encoding(Encoding::UTF_8) }
```

PR #482, opened 2024-09-04
Dup frozen columns array to force encoding in Trilogy

jenshenny:
Closes: https://github.com/ankane/blazer/issues/481

When using the Trilogy adapter and Rails 7.2, I'm getting a frozen array error when running a query. Columns is frozen in Rails 7.2 https://github.com/rails/rails/commit/9652207c6b12ccdfd7067b874f8ceff10c54d2b4. 

Instead of attempting to modify the array, changing to allocate a new one.


ankane:
Thanks @jenshenny!

PR #487, opened 2025-01-06
Add a new checks type: `new_bad_data`

Raymond-Youssef:
I needed an alert every time bad data occurred, hence this PR.
this new check type sends email notification/slack notification whenever more bad data is fetched.


ankane:
Hi @Raymond-Youssef, thanks for the PR. I'm not sure this is common enough to include, but I've added an experimental API for applications to add new check types in the commit above. Happy to review a PR that extends the API to support this check type.

PR #491, opened 2025-02-13
Introduce docs on SolidQueue setup

epugh:
I want to capture for other users how easy it is to use Blazer checks with SolidQueue (simpler than Cron!)....

I am still testing these docs.  I have confirmed `Blazer.send_failing_checks` works in my production Quepid setup.  http://app.quepid.com


ankane:
Hi @epugh, added in the commit above (note the `schedule` should match one of the keys in `blazer.yml`).

Issue #502, opened 2025-04-08
Forecasting with prophet-rb gives a `can't modify frozen Array` error


simonrentzke:
I'm not sure if I have not configured things correctly, but when I was trialing the use of this locally with this set:

`forecasting: prophet`

I get this error when trying to forecast on a small dataset:

`Error generating forecast: can't modify frozen Array`

<img width="1175" alt="Image" src="https://github.com/user-attachments/assets/b8b5278b-d953-494a-9922-f348a7f79a90" />

Rails 8, Ruby 3.2.2

ankane:
Fixed in #503

PR #503, opened 2025-04-08
Uses a dup of the frozen ActiveRecord columns so that forecasting can add an extra column

simonrentzke:
This resolves the problem mentioned here #502 

I'm not sure it's the correct solution though.

I suspect it's a similar problem that happened here https://github.com/ankane/blazer/pull/482


ankane:
Thanks @simonrentzke!

PR #512, opened 2025-12-02
Added support for in-statement variable annotations.

Jack12816:
First things first: thank you all very much for this awesome gem! :pray: 

**- What is it good for**

In my company and use case, the people who write and maintain queries in Blazer are not the ones running them, and neither of these groups maintains the Blazer configuration or deploys the application. Smart variables would be useful, but they require configuration changes and deployments. Simple variables are great, but they result in free-text inputs. The people running the queries are not always technical, so it would be helpful if we could annotate variable values so that, instead of free-text inputs, they see select boxes with valid options.

This PR adds a solution for this:

Screenshot_20251202_113229

**- What I did**

I just added a `Blazer::Statement#variable_values` method which searches for comments on the statement in a particular format, tries to parse them as JSON, and uses them for displaying on the UI. 

**- A picture of a cute animal (not mandatory but encouraged)**
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,6 +219,18 @@ Use `{start_time}` and `{end_time}` for time ranges. [Example](https://blazer.do
 SELECT * FROM ratings WHERE rated_at >= {start_time} AND rated_at <= {end_time}
 ```
 
+#### Variables with static values (via user selection)
+
+```sql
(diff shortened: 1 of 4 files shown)


marcusg:
+1 :+1:

PR #515, opened 2026-02-05
Added dark theme

joaomarceloods:
Added dark theme using `@media (prefers-color-scheme: dark)` to automatically match the user's device theme.

Moved existing (light) styles to `blazer.css` so `blazer-dark.css` can override selectors with dark styles.

Current light theme unchanged.

Screenshot 2026-02-04 at 21 07 29


ankane:
Hi @joaomarceloods, thanks for the PR, but it's not something I'd like to maintain.

PR #516, opened 2026-02-19
Add catalog support to Athena adapter

simonrentzke:
Allow setting a `catalog` in blazer.yml for Athena data sources to support non-default catalogs (e.g. S3 Table Catalog). Passes catalog to both Athena query_execution_context and Glue get_tables calls.


ankane:
Hi @simonrentzke, thanks for the PR! Merged a version of this in the commit above.

Issue #505, opened 2025-06-03
Variables in Checks


namolnad:
Hey @ankane, hope you've been well!

I've been adding a number of checks in our system lately, many of which are duplicated queries (warning systems for separate customers of ours), and it got me thinking about being able to use variables within checks. I wondered if it would be possible to add a jsonb column to the checks table, and allow users to create a new check referencing the same underlying query.

I figure it would be possible to go through the statement's variables and add fields for each of these in the outputted html when first creating a new check, and then store those as key-value pairs in the jsonb column for the check. Then [here](https://github.com/ankane/blazer/blob/e5ba950e0a9190d22654a33c9ff06caeb982c88c/lib/blazer/data_source.rb#L98) we could just inject those values before calling `.bind`. This would allow, e.g., reuse of a complex query that uses something like `where companies.id = {company_id}`, so that you could have separate checks for any company you're interested in being warned about separately.

I figured I'd surface the idea first to see what you thought before going too much further.

Thanks!

PR #522, opened 2026-06-22
docs: clarify variables in checks

dashitongzhi:
## Summary
- Clarifies that checks cannot use queries with variables
- Adds examples for converting variable-based queries into check-compatible queries
- Documents limitations and troubleshooting tips for scheduled checks

References #505.


ankane:
Hi @dashitongzhi, thanks for the PR, but I don't think it's needed.

PR #523, opened 2026-07-30
Avoid SassC evaluating CSS max() with custom properties

arlington6988:
This fixes SassC compilation errors when CSS custom properties are used inside native CSS math functions.

SassC attempts to evaluate `max()` as a Sass function and raises errors such as:

"var(--ts-pr-caret)" is not a number for `max`

This change preserves the intended browser behavior while avoiding SassC evaluation at build time.

This updates CSS `max()` calls that include CSS custom properties so they are emitted as native CSS functions instead of being evaluated by SassC at compile time.

### DIFF
--- a/app/assets/stylesheets/blazer/tom-select.css
+++ b/app/assets/stylesheets/blazer/tom-select.css
@@ -225,7 +225,7 @@
   cursor: pointer;
 }
 .plugin-clear_button.form-select .clear-button, .plugin-clear_button.single .clear-button {
-  right: max(var(--ts-pr-caret), 8px);
+  right: Max(var(--ts-pr-caret), 8px);
 }
 .plugin-clear_button.focus.has-items .clear-button, .plugin-clear_button:not(.disabled):hover.has-items .clear-button {
   opacity: 1;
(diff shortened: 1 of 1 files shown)


ankane:
Hi @arlington6988, the stylesheets aren't meant to be processed as Sass. Can you share the:

1. Rails version
2. Sass gem and version
3. Sass configuration
4. Full stack trace of the error

tvdeyen:
@ankane every Rails app still using sprockets and `sass-rails` (or its alias `sassc-rails`), which registers the Sass compiler. Sprockets stack traces are unfortunately pretty useless. This is all we got

```
Failure/Error:
   SassC::Engine.new(
     input,
     {
       style: :compressed
     }
   ).render

 ActionView::Template::Error:
   Error: "var(--ts-pr-caret)" is not a number for `max'
           on line 7066:10 of stdin, in function `max`
           from line 7066:10 of stdin
   >>   right: max(var(--ts-pr-caret), 8px);

      ---------^
 # stdin:7066
```

ankane:
Thank you both! Merged in 5d8b0a6f49164869e1e323a27a58cb2484916a70.

tvdeyen:
Thanks for addressing it so quickly

PR #524, opened 2026-07-31
Add missing sql aliases 

adamzapasnik:
After upgrading from 3.3 we're getting this error that is nor surfaced anywhere because `def tables` doesn't check errors and returns an empty array instead:
```
=> " subquery in FROM must have an alias\nLINE 1: SELECT * FROM (SELECT table_schema, table_name FROM informat...\n                      ^\nHINT:  For example, FROM (SELECT ...) [AS] foo.\n"
```

Tests are actually failing without this fix:

```
  1) Failure:
PostgresqlTest#test_tables_method [test/adapters/postgresql_test.rb:28]:
Expected [] to include "users".

  2) Failure:
PostgresqlTest#test_schema_method [test/adapters/postgresql_test.rb:40]:
--- expected
+++ actual
@@ -1 +1 @@
-[{name: "id", data_type: "bigint"}, {name: "name", data_type: "character varying"}]
+nil
```

### DIFF
--- a/lib/blazer/adapters/sql_adapter.rb
+++ b/lib/blazer/adapters/sql_adapter.rb
@@ -75,7 +75,7 @@ module Blazer
           if sqlite?
             "SELECT NULL, name FROM sqlite_master WHERE type IN ('table', 'view') ORDER BY name"
           elsif postgresql?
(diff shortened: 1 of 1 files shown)


ankane:
Thanks @adamzapasnik, it looks like this only affects Postgres 15 and below.

PR #528, opened 2026-07-31
Restore data-method link handling

### LINKED ISSUE #526: Delete links no longer work after removing rails-ujs
After upgrading Blazer from 3.3.0 to 3.5.0, the Delete button on the query edit page stopped working.

### Steps to reproduce

1. Use blazer 3.5.0 (or master).
2. Open any saved query, e.g. /blazer/queries/1-test.
3. Confirm the deletion.
4. The record is **not** deleted — the page just reloads/re-shows the same record.

Same problem reproduces for the Delete buttons on Dashboards and Checks.

### Expected behavior

The query is deleted.

### Actual behavior

After confirmation, the link is followed with a `GET` request instead of submitting a `DELETE` request:

```
Started GET "/blazer/queries/1-test" for ::1 at ...
Processing by Blazer::QueriesController#show as HTML
Parameters: {"id"=>"1-test"}
```

This appears to be a regression introduced by f51e2392aa7d2596b96f9186de452610f243fb54. The commit replaced the `data-confirm` behavior previously provided by `rails-ujs`, but did not replace the handling of `data-method`.

foxy-eyed:
Restores support for links with `data-method` after the removal of `rails-ujs`.

The new click handler creates and submits a temporary hidden form with the requested HTTP method and CSRF token, preserving the previous full-page navigation behavior without reintroducing the `rails-ujs` dependency.

It also handles `data-confirm` in the same listener, ensuring that cancelling the confirmation prevents the request from being submitted.

Tested locally against a Rails app using Turbo upgrading from blazer 3.3.0 → 3.5.0 (which surfaced the regression) — with this change, Delete correctly issues a `DELETE` request and removes the record.

Fixes #526

### DIFF
--- a/app/assets/javascripts/blazer/application.js
+++ b/app/assets/javascripts/blazer/application.js
@@ -54,12 +54,52 @@ document.addEventListener("click", function (e) {
 })
 
 document.addEventListener("click", function (e) {
-  const target = e.target.closest("a[data-confirm]")
-  if (target) {
-    if (!window.confirm(target.getAttribute("data-confirm"))) {
-      e.preventDefault()
-    }
+  const target = e.target.closest("a[data-confirm], a[data-method]")
+  if (!target) return
+
+  const confirmMessage = target.getAttribute("data-confirm")
+
+  if (confirmMessage && !window.confirm(confirmMessage)) {
+    e.preventDefault()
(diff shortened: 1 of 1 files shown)


ankane:
Thanks @foxy-eyed! Merged a version of this in the commit above.

